PR #5, opened 2020-11-17
rose_fileinstall to just fileinstall

wxtim:


setup.cfg
```diff
@@ -17,5 +17,5 @@
 [options.entry_points]
 cylc.configure = 
     configure = cylc.rose.rose:get_rose_vars
```
oliver-sanders, line 19: This name needs to be unique so best to name it `rose` or `cylc_rose` to avoid conflicts with any other plugins:

```suggestion
    rose = cylc.rose.rose:get_rose_vars
```

setup.cfg
```diff
@@ -17,5 +17,5 @@
...
-cylc.rose_fileinstall =
+cylc.fileinstall =
```
oliver-sanders, line 20: This entry point can be used for more than file installation, for example https://github.com/cylc/cylc-flow/issues/3780 would require installing packages into virtual environments.

```suggestion
cylc.post_install =
```

setup.cfg
```diff
@@ -17,5 +17,5 @@
...
     install = cylc.rose.rose:rose_fileinstall
```
oliver-sanders, line 21: Namespace the plugin's entry points to avoid possible conflicts:

```suggestion
    rose = cylc.rose.rose:rose_fileinstall
```

wxtim:
@oliver-sanders  - I think that I sorted the points that you mentioned.

oliver-sanders (approved):

PR #15, opened 2020-12-18
Jinja2 literal eval

### LINKED ISSUE #14: jinja2: interpretting values
## The Problem

At Cylc7/Rose2019 the `[jinja2:suite.rc]` variables were crudely pasted into a Jinja2 `{% set %}` statement. This meant that variables were written in the Jinja2 format which makes perfect sense.

However, with the addition of EmPy (another Python based templating engine) and more importantly the Python future of Cylc defining workflow variables in Python makes more sense.

As extra motives:

* Cylc7 does not have typed variables, Cylc8 does, we needed a format, Python makes the most sense, it opens the door to using things like Pandas dataframes or other Python objects in the construction of workflow definitions via a future Python configuration interface.
* The previous arrangement permits not just literals but also expressions which is highly undesirable.

## Literal Parsing

So the idea was to only parse Jinja2 vars from Rose configurations as Python literals since the two are very similar. Sadly, the exceptions to the rule appear to be much more plentiful than anticipated. Here follows a breakdown of the exceptions:

### Jinja2 Expressions

Unintended side-effect of the way Jinja2 vars were dumped into `{% set %}` statements.

#### Example:

```ini
[jinja2:suite.rc]
LIST=[0] + range(5) | list
INT=4 + (7 * 24)
```

#### Possible workaround:

None!

This is considered bad practice and something which should not have been possible in the first place.

### Multi-Line values

This refers to a subset of Jinja2 expressions that may be considered somewhat legitimate.

#### Examples:

```ini
[jinja2:suite.rc]
# 'foo bar baz'
MULTI_LINE_STRING='foo'
                 =' bar'
                 =' baz'
# [1,2,3]
MULTI_LINE_LIST=[
               =1,
               =2,
               =3,
               =]
```

#### Workaround:

`s|\n||g`

### Lower-Case Booleans

Jinja2 accepts lowercase booleans (which Python does not).

#### Example:

```ini
[jinja2:suite.rc]
PYTHONIC_BOOL=True
LOWERCASE_BOOL=true
```

#### Possible Workaround

`s/^true$/True/`, `s/^false$/False/`.

#### Example:

Jinja2 allows the declaration of tuples without the leading or trailing brackets:

```ini
[jinja2:suite.rc]
PYTHONIC_TUPLE=(1,2,3)
IMPLICIT_TUPLE=1,2,3
```

#### Workaround:

```python
try:
    evaluate_value(value)
except:
    # attempt to parse this as an implicit tuple
    value = (
        evaluate_value(item)
        for item in value.split(',')
    )
```

### Zero-Padded Integers

Take the value `0644`, this will be parsed as:

* `420` - Bash
* `644` - Jinja2
* `420` - Python2
* `SyntaxError` - Python3

Bash and Python2 both parse this as an octal, Jinja2 parses it as a decimal, Python3 requires either the removal of the leading `0` to parse it as a decimal *or* the addition of a leading `o` to parse it as an octal.

#### Example

```ini
[jinja2:suite.rc]
# harmless (in Python2/Bash)
DECIMAL=01
# not so harmless (in Python2/Bash)
DECIMAL=09
```

#### Workaround

Since Python2 octals are invalid Python3 octals it is possible to support the two side-by-side:

`s/^0+//`

## Rose Types

This highlights an additional issue with Rose.

It is possible to define an implicit tuple and make that editable in Rose Edit by setting the metadata `length`, this goes a long way to explaining why this approach has seen so much usage.

Rose does have a `python_list` type, but that does not allow you to specify the type of the data contained within in the manner that setting the `length` does.

Rose also lacks metadata types for:

* tuples
* dicts (not such an issue)
* sets (not an issue at Python2, slightly more of an issue at Python3)
* octals

oliver-sanders:
close #14 

Permit all valid Jinja2 literals.

Since Jinja2 literals are (mostly) a superset of Python literals this just broadens the scope of what's possible.

However, there is one quirk, Jinja2 does not support `{'python', 'set', 'notation'}`.

~Putting up as a draft due to an unrelated testing problem.~ Solved the crude way.


cylc/rose/jinja2_parser.py
```diff
@@ -0,0 +1,127 @@
...
+        ast = self.parse(templ)
+        stack = [ast]
+        while stack:
+            node = stack.pop()
+            if not isinstance(node, self._LITERAL_NODES):
+                raise Exception(
+                    f'Invalid literal: {value}'
+                    f'\n{type(node)}'
+                )
+            stack.extend([x for x in node.iter_child_nodes()])
```
wxtim, line 125: IMO this isn't very nice to read, but a few minutes looking at the jinja2 api docs and Stackoverflow have convinced me that there's no clear magic for doing this better. Jinja2 ast's look rather limited in their abilities.

Can I suggest that (depending on how balkanized you want the code) that you consider:
a) Moving this check into a function with a name such as `is_legal_jinja2(value)` or
b) The following slight refactor:
```suggestion

        # dump this value into a Jinja2 template
        # check that this expression consists only of literals
        templ = '{{ %s }}' % value
        ast = self.parse(templ)
        stack = [ast]
        while stack:
            node = stack.pop()
            if not isinstance(node, self._LITERAL_NODES):
                raise Exception(
                    f'Invalid literal: {value}'
                    f'\n{type(node)}'
                )
            stack.extend([x for x in node.iter_child_nodes()])

```

oliver-sanders: I don't think there is a much clearer way to express this code, it's very short and already has a comment:code ratio of 53:15.

oliver-sanders:
Still can't work out why adding `--doctest-modules` causes a Python import error, may have to cheat and remove the `pytest.ini` file :(

wxtim (approved):
I'm happy with this.
I'm going to create an issue to try and put the testing back. It's low priority, but ought to be recorded

cylc/rose/jinja2_parser.py
TomekTrzeciak: > I don't think there is a much clearer way to express this code, it's very short and already has a comment:code ratio of 53:15.

You could replace the stack with `Node.find_all` and avoid reparsing the expression, but other than that there isn't much fat that can be cut out here:

```python
        # dump this value into a Jinja2 template
        templ = '{{ %s }}' % value
        # check that this expression consists only of literals
        ast = self.parse(templ)
        for node in ast.find_all(jinja2.nodes.Node):
            if not isinstance(node, self._LITERAL_NODES):
                raise Exception(
                    f'Invalid literal: {value}'
                    f'\n{type(node)}'
                )
        # evaluate it
        return self.from_string(ast).render()  # <- avoids reparsing the expression
```

(Note that there is also [`jinja2.Environment.compile_expression`](https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.Environment.compile_expression) API but it fails some of your more perverse tests, namely the tuple ones, unless protected with extra parentheses: `return self.compile_expression('(%s)' % value)()`).

I wonder, however, if the following jinja-less solution wouldn't provide you with close enough approximation:

```python
        value = value.strip()
        if value in ('true', 'false'):
            value = value.capitalize()
        value = '(%s)' % value
        return ast.literal_eval(value)
```

This only fails on octal numbers (how widespread are these?), but it does work for sets.

Issue #9, opened 2020-12-10
catch rose errors and log them with Cylc


oliver-sanders:
Blocked by: https://github.com/metomi/rose/issues/2427

Cylc [catches all `cylc.CylcError` exceptions](https://github.com/cylc/cylc-flow/blob/a2f2a0d219d33207bc55f92a8f69273c2d532382/cylc/flow/terminal.py#L245-L255) and suppresses traceback to give the user a clear error message instead.

The cylc-rose repo should extend this functionality to Rose exceptions so that they also give clear error messages rather than pages of traceback.

One, crude, but effective way to do this is to patch Rose exceptions in the cylc-rose plugin to inherit from `CylcError`. Crude, but effective:

```python
>>> import cylc.flow.exceptions
>>> # this the exception to patch (e.g. from metomi.rose import ...)
>>> class Foo(Exception): pass
... 
>>> # this is the new exception which inherits both
>>> class FooError(cylc.flow.exceptions.CylcError, Foo): pass
... 
>>> # now we crudely patch it in :)
>>> cylc.flow.exceptions.CylcError = FooError
>>> # now to prove it works...
>>> e = cylc.flow.exceptions.CylcError('a')
>>> isinstance(e, CylcError)
True
>>> isinstance(e, FooError)
True
```

PR #16, opened 2020-12-30
catch literal_eval errors

oliver-sanders:
sibling: https://github.com/metomi/rose/pull/2430

Catch literal_eval errors and wrap them with `ConfigProcessError` to provide a more helpful error message.
Once #9 is done only the last line will appear in the Cylc command line output.

*Example:*

```ini
[jinja2:suite.rc]
X=Y
```

*Before:*

```
Traceback (most recent call last):
...
ValueError: malformed node or string: <_ast.Name object at 0x7f9c43c95ed0>
```

*After:*

```
Traceback (most recent call last):
...
metomi.rose.config_processor.ConfigProcessError: jinja2:suite.rc=Y=X: Invalid template variable: X
Must be a valid Python or Jinja2 literal (note strings "must be quoted").
```

### DIFF
--- a/cylc/rose/rose.py
+++ b/cylc/rose/rose.py
@@ -36,18 +36,22 @@ class MultipleTemplatingEnginesError(Exception):
     ...
 
 
-def get_rose_vars(dir_=None, opts=None):
-    """Load Jinja2 Vars from rose-suite.conf in dir_
+def get_rose_vars_from_config_tree(dir_=None, opts=None):
+    """Load template variables from Rose suite configuration.
+
+    This loads the Rose suite configuration tree from the filesystem
+    using the shell environment.
 
     Args:
-        dir_(string or Pathlib.path object):
-            Search for a ``rose-suite.conf`` file in this location.
+        dir_(string | pathlib.Path):
+            Path to the Rose suite configuration
+            (the directory containing the ``rose-suite.conf`` file).
         opts:
             Some sort of options object or string - To be used to allow CLI
             specification of optional configuaration.
 
     Returns:
-        A dictionary of sections of rose-suite.conf.
+        dict - A dictionary of sections of rose-suite.conf.
         For each section either a dictionary or None is returned.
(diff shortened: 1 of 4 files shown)


oliver-sanders:
(note: tests will fail until the sibling is merged)

cylc/rose/rose.py
```diff
@@ -35,18 +35,22 @@ class MultipleTemplatingEnginesError(Exception):
     ...
 
 
-def get_rose_vars(dir_=None, opts=None):
-    """Load Jinja2 Vars from rose-suite.conf in dir_
+def get_rose_vars_from_config_tree(dir_=None, opts=None):
```
wxtim, line 38: I think that you will need to change the entry point name in  `setup.cfg`.

wxtim (changes requested):
I think that you will need to change `setup.cfg` thus:
```diff
[options.entry_points]
cylc.pre_configure = 
-    rose = cylc.rose.rose:get_rose_vars
+    rose = cylc.rose.rose:get_rose_vars_from_config_tree
```

cylc/rose/rose.py
```diff
@@ -73,18 +78,51 @@ def get_rose_vars(dir_=None, opts=None):
...
+
+    # Export environment vars
+    for key, val in config['env'].items():
+        os.environ[key] = val
+
+    return config
+
+
+# pre_configure entry point
+get_rose_vars = get_rose_vars_from_config_tree
```
oliver-sanders, line 96: Au contraire.

wxtim: Missed that.
Happy now.

wxtim (approved):

cylc/rose/rose.py
oliver-sanders: Bit of a strange thing to do but I thought I'd save you a `pip install`.

datamel (approved):

PR #20, opened 2021-01-05
[Test failure may be OK] Create cylc config save file

wxtim:
sibling: https://github.com/cylc/cylc-flow/pull/4023

### Summary
Partial response to [cylc-flow:#3819](https://github.com/cylc/cylc-flow/issues/3819)
Required by a future PR to cylc-flow adding command line options similar to the following `rose suite-run` options:
- `--opt-conf-key`
- `--defines`
- `--suite-defines`

Some auto tests will fail if Cylc-flow is master.

### Checklist
- [x] Written functionality.
- [x] Unit-ized functionality.
- [x] Written what I think are fairly thorough unit tests for non-top level functions.
- [x] Written a basic set of functional tests for the top level function.
    - [x] Files are produced in the expected way.
    - [x] Output nodes are what you'd expect.
- [x] Read through and double checked.
- [x] Re-read (proposal)[https://github.com/cylc/cylc-admin/blob/master/docs/proposal-cylc-rose-installing-rose-configs.md].
- [x] Sure that there are enough functional tests: I think that there are enough unit tests that functional testing can be quite a small selection. I've lifted these cases more or less wholesale from the [cylc install proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-cylc-rose-installing-rose-configs.md).
- [x] Manually check coverage of new code. - 100% for new code. [rose.py.pdf](https://github.com/cylc/cylc-rose/files/5820147/rose.py.pdf)

### Outstanding questions

[_cylc install proposal_](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-cylc-rose-installing-rose-configs.md)

Is this the `rose-suite-cylc-install.conf` file. Does this need doing in `cylc/cylc-flow`?


cylc/rose/rose.py
```diff
@@ -255,3 +259,125 @@ def rose_config_tree_loader(dir_=None, opts=None):
...
+    """
+    # Construct new ouput based on optional Configs
+    newconfig = []
+    newconfig = ConfigNode()
+    # Currently unclear how Rose deas with these.
+    # Coming back late because treatent of args more obvious.
+    if opt_conf_keys:
+        newconfig.set(
+            ["opts"], ' '.join(opt_conf_keys), ConfigNode.STATE_USER_IGNORED
+        )
```
oliver-sanders, line 297: Quick summary of the proposal with relation to opt configs passed in on the CLI:

**1. In the event of duplicates, remove the earliest occurrences.**

Here's an example algorithm that does the reverse (removes the latest occurrences):

```python
def simplify_path(path):
    """Removes duplication in paths whilst maintaining integrity.

    If duplicate items are present in a path this keeps the first item and
    removes any subsequent duplicates.

    Examples:
        >>> simplify_path('')
        ''
        >>> simplify_path('a')
        'a'
        >>> simplify_path('a:a:a')
        'a'
        >>> simplify_path('a:b:a')
        'a:b'
        >>> simplify_path('a:b:b:a')
        'a:b'
        >>> simplify_path('a:b:a:b:c:d:a:b:c:d:e')
        'a:b:c:d:e'

    """
    path = path.split(':')
    counter = Counter(path)
    for item, count in counter.items():
        ptr = len(path) - 1
        while count > 1:
            if path[ptr] == item:
                path.pop(ptr)
                count -= 1
            ptr -= 1
    return ':'.join(path)
```

**2. Add these confs to the `cylc-install` opt config as an ignored variable.**

**3. Append these opts to the end of the list in `rose-suite.conf|opts`.**

wxtim: Thanks.

oliver-sanders:
I think this should close #3819 wdyt?

wxtim:
I didn't enjoy rebasing this - I think that there is scope for tidying up the testing a little before release, but if things are functional I should like to have this PR merged.

oliver-sanders:
> Is this the rose-suite-cylc-install.conf file. Does this need doing in cylc/cylc-flow?

Yes, and yes.

This needs adding to the `cylc-install` rsync exclude list.

oliver-sanders:
`cylc install` is erroring because the `opts` kwarg is not always passed through to the plugin and the code doesn't check `is opts` before trying to access attributes of `opts`.

https://github.com/cylc/cylc-flow/pull/4023#issuecomment-762782107

tests/test_record_cylc_install_options.py
```diff
@@ -0,0 +1,186 @@
...
+            # Expected rose-suite.conf (Only need opts section).
+            'opts=a b c d'
+        )
+    ]
+)
+def test_functional_record_cylc_install_options(
+    monkeypatch, tmp_path, opts, rose_suite_conf, install_file, env_inserts,
+    kgo_install_file, kgo_rose_file
+):
+    class MockConfigLoader(ConfigLoader):
```
wxtim, line 139: @oliver-sanders  I have become concerned that given that this test is basically a functional test of one of the outward facing functions that this mocking approach will become very messy once we start adding (for example) `log` files.
Is it OK if I roll back on it and just touch the file system.

oliver-sanders: In order to use the rose Config API would need to pull the CLI and API parts into separate functions.

Here's a quick attempt, I took the `record_cylc_install_options` function and re-ordered the different bits like so:

* load from FS
* process
* dump to FS

The process bit is the bit you want to test.

```python
def record_cylc_install_options(
    dest_root=None,
    opts=None,
    dir_=None,
):
    # LOAD FILES
    opt_conf_filepath = Path(dest_root) / 'rose-suite-cylc-install.conf'
    rose_conf_filepath = Path(dest_root) / 'rose-suite.conf'
    dumper = ConfigDumper()
    loader = ConfigLoader()

    new_opt_config = get_cli_opts_node(opts)

    if opt_conf_filepath.is_file():
        old_opt_config = loader.load(str(opt_conf_filepath))
    else:
        old_opt_config = ConfigNode()

    if not rose_conf_filepath.is_file():
        rose_conf_filepath.touch()
    rose_suite_conf = loader.load(str(rose_conf_filepath))

    # PROCESS FILES
    new_opt_config, rose_suite_conf = _record_cylc_install_options(
        old_opt_config,
        new_opt_config,
        rose_suite_conf
    )

    # DUMP FILES
    dumper.dump(new_opt_config, str(opt_conf_filepath))

    return new_opt_config, rose_suite_conf['opts']


def _record_cylc_install_options(
    old_opt_config,
    new_opt_config,
    rose_suite_conf
):
    # THIS HAS NO FILESYSTEM LOGIC SO IS GOOD FOR TESTING
    if old_opt_config:  # TODO check this does what you want
        new_opt_config = merge_rose_cylc_suite_install_conf(
            old_opt_config,
            new_opt_config
        )
        new_opt_config.comments = [
            ' This file records CLI Options.'
        ]
    rose_suite_conf = get_installed_rose_suite_conf_node(
        rose_suite_conf,
        new_opt_config
    )
    return new_opt_config, rose_suite_conf
```

wxtim:
@oliver-sanders  and @datamel  Ready for review. Sorry about all the messing about.

cylc/rose/entry_points.py
```diff
@@ -0,0 +1,207 @@
...
+            Path to dump the rose-suite-cylc-conf
+
+    Returns:
+        cli_config - Config Node which has been dumped to
+        ``rose-suite-cylc-install.conf``.
+        rose_suite_conf['opts'] - Opts section of the config node dumped to
+        installed ``rose-suite.conf``.
+    """
+    # Construct a path objects representing our target files.
+    conf_filepath = Path(dest_root) / 'rose-suite-cylc-install.conf'
```
oliver-sanders, line 176: To make this register with Rose as an opt config it needs to go into the `opt/` dir.

```suggestion
    conf_filepath = Path(dest_root, 'opt', 'rose-suite-cylc-install.conf'
```

cylc/rose/entry_points.py
```diff
@@ -0,0 +1,207 @@
...
+
+    if conf_filepath.is_file():
+        oldconfig = loader.load(str(conf_filepath))
+        cli_config = merge_rose_cylc_suite_install_conf(oldconfig, cli_config)
+    dumper(cli_config, str(conf_filepath))
+
+    cli_config.comments = [
+        ' This file records CLI Options.'
+    ]
+    dumper.dump(cli_config, str(conf_filepath))
```
oliver-sanders, line 195: Is this dumping the config twice? Once without the comment and once with?

cylc/rose/entry_points.py
```diff
@@ -0,0 +1,207 @@
...
+    # Create a config based on command line options:
+    cli_config = get_cli_opts_node(opts)
+
+    # If file exists we need to merge with our new config, over-writing with
+    # new items where there are duplicates.
+
+    if conf_filepath.is_file():
+        oldconfig = loader.load(str(conf_filepath))
+        cli_config = merge_rose_cylc_suite_install_conf(oldconfig, cli_config)
+    dumper(cli_config, str(conf_filepath))
```
oliver-sanders, line 190: Would be good to avoid writing this file if no options were set.

I think we should be able to do:

```
if cli_config:
```

or:

```
if cli_config.values():
```

cylc/rose/entry_points.py
```diff
@@ -0,0 +1,207 @@
...
+    # Replace the opts section of the rose-suite.conf in the install location.
+    # If we have not a rose-suite.conf but we use one of the Rose-style
+    # options we still want to record those options.
+    if not rose_conf_filepath.is_file():
+        rose_conf_filepath.touch()
+    rose_suite_conf = loader.load(str(rose_conf_filepath))
+    rose_suite_conf = get_installed_rose_suite_conf_node(
+        rose_suite_conf, cli_config
+    )
+    dumper(rose_suite_conf, rose_conf_filepath)
```
oliver-sanders, line 206: Need to fiddle the `opts` value to insert the `cylc-install` optional config (if one was written) otherwise subsequent Rose commands will not load it.

cylc/rose/entry_points.py
wxtim: I've misunderstood this then. I didn't realize that it was meant to go in the opts folder. That makes much more sense now.

cylc/rose/entry_points.py
wxtim: Looks like I've made a bit of a mess here. I'll go look at this in detail.

cylc/rose/entry_points.py
wxtim: See my comments about misunderstanding how this is meant to work.

cylc/rose/entry_points.py
wxtim: Now the whole function exits earlier if there are no cli options to work with...

cylc/rose/entry_points.py
wxtim: Looks like it. What a mess.

cylc/rose/entry_points.py
```diff
@@ -0,0 +1,208 @@
...
+                'env': {'MYVAR': 42},
+                'empy:suite.rc': None,
+                'jinja2:suite.rc': {
+                    'myJinja2Var': {'yes': 'it is a dictionary!'}
+                }
+            }
+
+    TODO:
+        - Consider allowing ``[jinja2:flow.conf]`` as an alias for
+          consistency with cylc.
```
oliver-sanders, line 69: Convert to issue?

wxtim: https://github.com/cylc/cylc-rose/issues/25

oliver-sanders:
> 1. Any new “defines” should be written to a new optional configuration with a special protected name

✅ Tested `-D` and `-S`.
✅ The `(cylc-install)` config is added to the end of the list.

> 2. The cylc reinstall command should always move the cylc-install opt conf to the end of the list

⌛ Will test the reinstall part once we have a reinstall command.

> 3. The cylc reinstall command should have an option to delete the cylc-install opt conf

⌛ Saving this one for a future PR.

> 4. cylc install should record optional configs to the cylc-install optional config

❌ via `-O` CLI opt.
✅ via `ROSE_SUITE_OPT_CONF_KEYS` env var.
❌ `cylc-install|opts` are appended to `rose-suite.conf|opts`


my test suite:

```console
$ find .  -type f -name '*.conf' -exec echo -e '\n=== {} ===' \; -exec cat '{}' \;

=== ./opt/rose-suite-foo.conf ===
[jinja2:suite.rc]
Y="f"

=== ./opt/rose-suite-bar.conf ===
[jinja2:suite.rc]
Y="b"

=== ./rose-suite.conf ===
opts=

[jinja2:suite.rc]
Y="base"
```

my test install:

```console
$ ROSE_SUITE_OPT_CONF_KEYS=foo cylc install -O bar -D '[env]a=b' -S 'a="b"'
INSTALLED foo from ~/roses/foo -> ~/cylc-run/foo/run1
```

the result:

```console
$ find ~/cylc-run/foo/runN/ -type f -name '*.conf' -exec echo -e '\n=== {} ===' \; -exec cat '{}' \;

=== ~/cylc-run/foo/runN//opt/rose-suite-foo.conf ===
[jinja2:suite.rc]
Y="f"

=== ~/cylc-run/foo/runN//opt/rose-suite-bar.conf ===
[jinja2:suite.rc]
Y="b"

=== ~/cylc-run/foo/runN//opt/rose-suite-cylc-install.conf ===
# This file records CLI Options.

!opts=foo

[env]
a=b

[jinja2:suite.rc]
a="b"

=== ~/cylc-run/foo/runN//rose-suite.conf ===
# Config Options '(cylc-install)' from CLI appended to options already in `rose-suite.conf`.
!opts=(cylc-install)

[jinja2:suite.rc]
Y="base"
```

the diff:

```diff
 === ~/cylc-run/foo/runN//opt/rose-suite-cylc-install.conf ===
 # This file records CLI Options.

-!opts=foo
+!opts=bar foo

 [env]
 a=b

 [jinja2:suite.rc]
 a="b"

 === ~/cylc-run/foo/runN//rose-suite.conf ===
 # Config Options '(cylc-install)' from CLI appended to options already in `rose-suite.conf`.
-!opts=(cylc-install)
+!opts=bar foo (cylc-install)

 [jinja2:suite.rc]
 Y="base"
```

oliver-sanders:
`rose-suite.conf|opts` setting is still ignored (which means it won't be loaded by `metomi.rose.config_tree`).

cylc/rose/utilities.py
```diff
@@ -0,0 +1,441 @@
...
+        cli_opts = ''
+    if 'opts' not in rose_conf:
+        rose_conf.set(['opts'], '')
+    rose_conf['opts'].comments = [(
+        f' Config Options \'{cli_opts} (cylc-install)\' from CLI appended to '
+        'options already in `rose-suite.conf`.'
+    )]
+    rose_conf_opts_str = ' '.join(
+        f'{rose_conf["opts"].value} {cli_opts} (cylc-install)'.split()
+    )
```
oliver-sanders, line 351: Just spotted, if the `rose-suite.conf` has configured `opts` as `!` "ignored" AND they have provided opt configs on the CLI then I think we just ignore the `!` "ignored" value and use the CLI opts i.e:

```suggestion
    if rose_conf['opts'].state in ['!', '!!']:
        rose_conf_opts_str = cli_opts.split() + ['(cylc-install)']
    else:
        rose_conf_opts_str = ' '.join(
            f'{rose_conf["opts"].value} {cli_opts} (cylc-install)'.split()
        )
```

setup.cfg
```diff
@@ -16,6 +16,7 @@
 
 [options.entry_points]
 cylc.pre_configure = 
-    rose = cylc.rose.rose:get_rose_vars
+    rose = cylc.rose.entry_points:get_rose_vars
 cylc.post_install =
-    rose = cylc.rose.rose:rose_fileinstall
+    rose_opts = cylc.rose.entry_points:record_cylc_install_options
+    rose_files = cylc.rose.entry_points:rose_fileinstall
```
oliver-sanders, line 22: The ability to use environment variables to control file installation has been broken again. Here's my test suite:

```
$ tree
.
├── base
├── flow.cylc
├── opt
│   └── rose-suite-x.conf
├── rose-suite.conf
└── x
```

```console
$ cat rose-suite.conf 
[env]
FILENAME=base

[file:myfile]
source=localhost:$PWD/$FILENAME
```

```console
$ cat opt/rose-suite-x.conf 
[env]
FILENAME=x
```

This should install a file called `myfile` which will either contain the contents of file `base` or file `x` depending on whether the `x` optional config is specified on the CLI.

```console
$ cylc install
...
PluginError: Error in plugin cylc.post_install.rose_files
file:myfile=source: FILENAME: unbound variable
```

This seems to be repeatable, it might be because the plugins are not necessarily called in the order that they are listed in the `setup.cfg`. I think it would be safer to group these two into a single entry point in order to ensure that they will always get run in this order, e.g. with a simple dumb wrapper like so:

```python
def whatever(*args, **kwargs):
    record_cylc_install_options(*args, **kwargs)
    rose_fileinstall(*args, **kwargs)
```

wxtim: >  it might be because the plugins are not necessarily called in the order that they are listed

The plugin which _should_ export `$FILENAME` to the environment is the pre-install plugin, so it should _always_ be run before the other two. So I don't think this is likely. More likely, somehow is environment issues:
Running `FILENAME=x cylc install` works. I wonder if the entry points run in their own envs.

cylc/rose/entry_points.py
```diff
@@ -0,0 +1,204 @@
...
+        from metomi.rose.fs_util import FileSystemUtil
+
+        # Update config tree with install location
+        # NOTE-TO-SELF: value=os.environ["CYLC_SUITE_RUN_DIR"]
+        config_tree.node = config_tree.node.set(
+            keys=["file-install-root"], value=str(dest_root)
+        )
+
+        # Artificially set rose to verbose.
+        # TODO - either use Cylc Log as event handler, or get Cylc Verbosity
```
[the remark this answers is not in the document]
wxtim: https://github.com/cylc/cylc-rose/issues/27

cylc/rose/entry_points.py
```diff
@@ -0,0 +1,204 @@
...
+        # settings to pass to Rose Reporter.
```
wxtim, line 122:
```suggestion
```

oliver-sanders:
```
$ cylc install
...
PluginError: Error in plugin cylc.post_install.rose_files
can only concatenate str (not "bytes") to str
```

setup.cfg
oliver-sanders: I'm not aware of any gaurentee that entry points will be called in the order they are specified in `setip.cfg` so yep, good to try a wrapper.

oliver-sanders:
cylc-flow: tim/cli-setting-of-rose-opts@d7f6cb18214776f63a336af1f478f584e1827192
cylc-rose: tim/create-cylc-config-save-file@6e7e0172bcaba225dc4961406f89e661d7c6853a

```
$ cylc install
...
PluginError: Error in plugin cylc.post_install.rose_files
can only concatenate str (not "bytes") to str
```

wxtim:
> cylc-flow: tim/cli-setting-of-rose-opts@d7f6cb18214776f63a336af1f478f584e1827192
> cylc-rose: tim/create-cylc-config-save-file@6e7e0172bcaba225dc4961406f89e661d7c6853a
> 
> ```
> $ cylc install
> ...
> PluginError: Error in plugin cylc.post_install.rose_files
> can only concatenate str (not "bytes") to str
> ```

I just can't replicate this @oliver-sanders

oliver-sanders:
Here's an example flow:

```
$ find . -type f -exec echo -e '\n# {}\n' \; -exec cat '{}' \;

# ./flow.cylc


# ./opt/rose-suite-x.conf

[env]
FILENAME=x

# ./rose-suite.conf

[env]
FILENAME=base

[file:myfile]
source=localhost:$PWD/$FILENAME

# ./x

x

# ./base

base
```

oliver-sanders:
Opt conf key stuff looks good, for interest here is my test:

```console
$ lsd
# ./opt/rose-suite-y.conf
[env]
O=Y

# ./opt/rose-suite-x.conf
[env]
O=X

# ./suite.rc

# ./rose-suite.conf
[env]
O=B


# opt conf overrides base conf
$ cylc clean conf-test; cylc install -O x
$ rose config --file ~/cylc-run/conf-test/runN/rose-suite.conf env O
X

# ROSE_SUITE_OPT_CONF_KEYS override base conf in the correct order
$ cylc clean conf-test; ROSE_SUITE_OPT_CONF_KEYS='x y' cylc install
$ rose config --file ~/cylc-run/conf-test/runN/rose-suite.conf env O
Y

# -O applied after ROSE_SUITE_OPT_CONF_KEY
$ cylc clean conf-test; ROSE_SUITE_OPT_CONF_KEYS='x y' cylc install -O x
$ rose config --file ~/cylc-run/conf-test/runN/rose-suite.conf env O
X
```

oliver-sanders:
> 1. Any new “defines” should be written to a new optional configuration with a special protected name

✅ Tested `-D` and `-S`.
✅ The `(cylc-install)` config is added to the end of the list.

> 2. The cylc reinstall command should always move the cylc-install opt conf to the end of the list

⌛ Will test the reinstall part once we have a reinstall command.

> 3. The cylc reinstall command should have an option to delete the cylc-install opt conf

⌛ Saving this one for a future PR.

> 4. cylc install should record optional configs to the cylc-install optional config

✅ via `-O` CLI opt.
✅ via `ROSE_SUITE_OPT_CONF_KEYS` env var.
✅ `cylc-install|opts` are appended to `rose-suite.conf|opts`

oliver-sanders (changes requested):
All good except one last thing :(

The `.rose-config_processors-file.db` gets created in the source directory rather than the run directory.

This means that file installation from one installation can effect file installation in another.

It may be necessary to `cd` into the run dir before the file installation then back again afterwards (perhaps in a try/finally).

wxtim:
> The .rose-config_processors-file.db gets created in the source directory rather than the run directory.
I believe that I have now fixed this. :)

cylc/rose/entry_points.py
```diff
@@ -103,28 +103,35 @@ def rose_fileinstall(dir_=None, opts=None, dest_root=None):
...
+
+            # Artificially set rose to verbose.
+            event_handler = Reporter(3)
+            fs_util = FileSystemUtil(event_handler)
+            popen = RosePopener(event_handler)
+
+            # Process files
+            config_pm = ConfigProcessorsManager(event_handler, popen, fs_util)
+            config_pm(config_tree, "file")
+        os.chdir(startpoint)
```
oliver-sanders, line 134: Ideally should stick this in a finally block just to be safe.

wxtim:
```suggestion
            os.chdir(startpoint)
```

cylc/rose/entry_points.py
```diff
@@ -0,0 +1,209 @@
...
+
+            # Artificially set rose to verbose.
+            event_handler = Reporter(3)
+            fs_util = FileSystemUtil(event_handler)
+            popen = RosePopener(event_handler)
+
+            # Process files
+            config_pm = ConfigProcessorsManager(event_handler, popen, fs_util)
+            config_pm(config_tree, "file")
+            os.chdir(startpoint)
```
wxtim, line 134:
```suggestion
    finally:
        os.chdir(startpoint)

oliver-sanders:
Good to go here, lets get https://github.com/cylc/cylc-flow/pull/4023 in.

oliver-sanders (approved):
@datamel could you review 2.

datamel (approved):

Issue #12, opened 2020-12-11
functional tests


oliver-sanders:
Test the cylc-rose functionality from within the cylc-rose repository (rather than in cylc-flow).

There are reasons why this should be done here:

* Nicer (less stuff in cylc-flow).
* Ensures we have the correct Cylc/Rose version bracketing.
* Allows us to have a testing matrix with different Rose/Cylc versions to ensure inter-compatibility.

Suggested approach, use pytest, avoid the filesystem where possible:

Split the existing functions into two to separate the cylc interface and fileparsing logic from the cylc-rose internals e.g:

```python
# the entry point
def get_rose_vars(dir_, opts):
    # illustrative and very non-functional :/
    config_tree = rose_config_tree_loader(dir_, opts)
    node = config_tree.node
    vars = _get_rose_vars(node)
    # ...
    config = {}
    config['whatever'] = vars['whatever']
    # ...
    return {'whatever': vars}

def _get_rose_vars(node):
    # ...
```

Then test the internal cylc-rose bits without the cylc-flow interface overheads and associated filesystem meddling:

```python
def test_env():
    # illustrative and very non-functional :/
    node = ConfigNode()
    node.set(['env', 'foo'], '1')
    assert _get_rose_vars(node) == {
        'whatever': {
             'foo' 1
        }
    }
```

wxtim:
I now have a branch https://github.com/wxtim/cylc-rose/tree/rose-fileinstall-recording.convert-functional-tests where all but one of the functional tests have been moved to rose-cylc. This branch is, however, a branch of my post install feature branch.

PR #24, opened 2021-01-18
Ignore ROSE_VERSION and CYLC_VERSION

### LINKED ISSUE #11: ignore ROSE_VERSION/CYLC_VERSION in rose-suite.conf
#### tldr

Exclude any `ROSE_VERSION` and `CYLC_VERSION` environment variables set in the `rose-suite.conf` from being exported into the environment, messing with templatvars or whatever.

#### context

In Rose2019 there was some level of support for the `[env]ROSE_VERSION` and `[env]CYLC_VERSION` variables.

If `ROSE_VERSION` was set and did not match `rose version` then rose exited with an error.

If `CYLC_VERSION` was set Rose exported it to `cylc` executable calls (to pick up the correct version via the wrapper script).

Now that cylc-rose is a plugin and Rose2/Cylc8 are installed in the same Python environment this system is meaningless and no longer functions.

However, many Rose2019 suites will have these variables defined, they have the potential to mess with the "true" `*_VERSION` variables and cause problems.

wxtim:
Closes #11 
Doesn't create any further tech debt, and provides a template for #12 

### Summary
Prevent user setting values for `CYLC_VERSION` and `ROSE_VERSION` from rose config.
- `ROSE_VERSION` will be retrieved from `metomi.rose.__version__`. If the user had set it otherwise a warning will be printed.
- `CYLC_VERSION` will be deleted with a warning.

- [x] Addressed issue
- [x] Created a new test to test this fix.

### DIFF
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -91,6 +91,24 @@ def get_rose_vars_from_config_node(config, config_node, environ):
         if rose_orig_host:
             config_node[section].set(['ROSE_ORIG_HOST'], rose_orig_host)
 
+        # Remove CYLC_VERSION to prevent conflict with Cylc 8.
+        if 'CYLC_VERSION' in config_node[section]:
+            LOG.warning(
+                f'CYLC_VERSION={config_node[section]["CYLC_VERSION"].value} '
+                'in rose-suite.conf will be ignored '
+                'to prevent conflict with Cylc 8 settings.'
+            )
+            del config_node[section]['CYLC_VERSION']
+        # If ROSE_VERSION is not the version we are using warn and replace.
+        if 'ROSE_VERSION' in config_node[section]:
+            LOG.warning(
+                'ROSE_VERSION='
+                f'{config_node[section]["ROSE_VERSION"].value} '
+                'in rose-suite.conf will be ignored. You are '
+                f'using version {ROSE_VERSION}.'
+            )
+        config_node[section].set(['ROSE_VERSION'], ROSE_VERSION)
+
         # Use env_var_process to process variables which may need expanding.
         for key, node in config_node.value[section].value.items():
             try:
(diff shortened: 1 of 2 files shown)


cylc/rose/rose.py
```diff
@@ -143,9 +145,27 @@ def get_rose_vars_from_config_node(config, config_node, environ):
...
+            del config_node[section]['CYLC_VERSION']
+        # If ROSE_VERSION is not the version we are using warn and replace.
+        if 'ROSE_VERSION' in config_node[section]:
+            if config_node[section]['ROSE_VERSION'] != ROSE_VERSION:
+                LOG.warning(
+                    'ROSE_VERSION='
+                    f'{config_node[section]["ROSE_VERSION"].value} '
+                    f'replaced with {ROSE_VERSION}: That is the version you '
+                    'are using.'
+                )
```
MetRonnie, line 166: And for this one, I would go for something along the lines of:
```suggestion
            LOG.warning(
                'ROSE_VERSION='
                f'{config_node[section]["ROSE_VERSION"].value} '
                'in rose-suite.conf will be ignored. You are '
                f'using version {ROSE_VERSION}.'
            )
```
I would get rid of the `if`; otherwise, if someone puts `ROSE_VERSION=2`, they won't get a warning and won't know not to do this

wxtim:
Thank you for suggestions Ronnie. I always find it really hard craft useful error messages while I am writing code. I think I'm generally too close to the problem.

MetRonnie (approved):

wxtim:
@MetRonnie  - I've re-triggered a review because I've done an annoying rebase and would appreciate another check.

MetRonnie (approved):

wxtim:
This is now unpleasantly conflicted with the base branch...

wxtim:
Closing this for now, since the issue is still undecided. Will wait for question to be agreed.

PR #28, opened 2021-02-08
Rose fileinstall recording.convert functional tests

### LINKED ISSUE #12: functional tests

wxtim:
Closes #12 within this repo. A companion Cylc PR is still required.

Move functional tests for cylc-rose currently in the Cylc-flow repo's prove based framework to using Pytest in this repo.


tests/functional/04_opts_set_from_env/04-opts-set-from-env/flow.cylc
```diff
@@ -0,0 +1,36 @@
+#!jinja2
```
oliver-sanders, line 1: Is this meant to have `04_opts_set_from_env` in the path twice?

(same for some other files)

tests/functional/test_pre_configure.py
```diff
@@ -0,0 +1,111 @@
...
+    sub = subprocess.run(
+        ['cylc', 'validate', str(srcdir)], capture_output=True
+    )
+
+    assert sub.returncode != 0
+    if expect:
+        assert sub.stderr == (
+                "FileParseError: Plugins set templating engine = empy which "
+                "does not match #!jinja2 set in flow.cylc."
+            )
```
oliver-sanders, line 48: Unused?

oliver-sanders (commented):
Tests seem unhappy.

tests/functional/04_opts_set_from_env/04-opts-set-from-env/flow.cylc
wxtim: No. This is a mistake.

wxtim: Can you see this when browsing the source? I can't, either on GH or locally!

tests/functional/test_pre_configure.py
wxtim: It is. It shouldn't be.

oliver-sanders (approved):

datamel (approved):

PR #29, opened 2021-02-08
Create cylc config save file.create log rose conf

### LINKED ISSUE #10: post_install: dump the parsed configuration to conf/<timestamp>-rose-suite.conf
https://cylc.github.io/cylc-admin/proposal-rose-suite-run.html#log-file-changes

wxtim:
Partially Closes #10 

Dump a record of the `rose-suite.conf` used as `log/<timestamp>-rose-suite.conf`.


cylc/rose/entry_points.py
```diff
@@ -98,6 +95,9 @@ def rose_fileinstall(dir_=None, opts=None, dest_root=None):
         dest_root (string or pathlib.Path)
 
     """
+    dir_ = pathchecker(dir_)
+    dest_root = pathchecker(dest_root)
```
oliver-sanders, line 99: Can do `Path(dest_root)` (Path accepts Path objects as arguments).

cylc/rose/entry_points.py
```diff
@@ -134,6 +134,9 @@ def rose_fileinstall(dir_=None, opts=None, dest_root=None):
         finally:
             os.chdir(startpoint)
 
+    from cylc.rose.utilities import dump_rose_log
```
oliver-sanders, line 137: Have already imported from `cylc.rose.utilities` at the top of the file so no performance improvement to doing it later on.

cylc/rose/utilities.py
```diff
@@ -442,3 +442,31 @@ def simplify_opts_strings(opts):
             seen_once.append(item)
 
     return ' '.join(reversed(seen_once))
+
+
+def dump_rose_log(dest_root, node):
+    from metomi.rose.config import ConfigDumper
+    from metomi.isodatetime.datetimeoper import DateTimeOperator
```
oliver-sanders, line 449: Have already imported `metomi.rose.config` at the top of the file, fine to move these both up there.

tests/test_config_node.py
```diff
@@ -94,3 +96,20 @@ def test_add_cylc_install_to_rose_conf_node_opts(rose_conf, cli_conf, expect):
         'already in `rose-suite.conf`.'
     )]
     assert result.state == ''
+
+
+def test_dump_rose_log(monkeypatch, tmp_path):
+    # Pin down the results of the function used to provide a timestamp.
+    def fake(*arg, **kwargs):
+        return '18151210T0000Z'
```
oliver-sanders, line 104: I lovelace it.

tests/test_config_node.py
```diff
@@ -94,3 +96,20 @@ def test_add_cylc_install_to_rose_conf_node_opts(rose_conf, cli_conf, expect):
...
+        DateTimeOperator, 'process_time_point_str', fake
+    )
+
+    node = ConfigNode()
+    node.set(['opts'], 'A B C')
+
+    dump_rose_log(tmp_path, node)
+
+    assert (tmp_path / 'log/18151210T0000Z-rose-suite.conf').read_text() == \
+        'opts=A B C\n'
```
oliver-sanders, line 115: Curious as the dumper strips the `opts` in normal usage (expected), perhaps test with `[env]`.

cylc/rose/entry_points.py
wxtim: I thought so, but had a few problems getting this working. Fixed now.

cylc/rose/entry_points.py
wxtim: And because I'm having that sort of day I alphabetized the imports too.

tests/test_config_node.py
wxtim: Do you want both, or `[env]` instead?

oliver-sanders: I think just `[env]` because I don't think we would expect this example to work from the CLI?

cylc/rose/entry_points.py
MetRonnie: Still needs to be moved to the top?

cylc/rose/utilities.py
```diff
@@ -442,3 +439,15 @@ def simplify_opts_strings(opts):
             seen_once.append(item)
 
     return ' '.join(reversed(seen_once))
+
+
+def dump_rose_log(dest_root, node):
```
MetRonnie, line 444: Doctsring (and/or type annotations would be good here)

tests/test_config_node.py
```diff
@@ -94,3 +96,20 @@ def test_add_cylc_install_to_rose_conf_node_opts(rose_conf, cli_conf, expect):
...
     )]
     assert result.state == ''
+
+
+def test_dump_rose_log(monkeypatch, tmp_path):
+    # Pin down the results of the function used to provide a timestamp.
+    def fake(*arg, **kwargs):
+        return '18151210T0000Z'
+    monkeypatch.setattr(
+        DateTimeOperator, 'process_time_point_str', fake
```
MetRonnie, line 106: A compact way of doing this would be `lambda *a, **k: '18151210T0000Z'` instead of defining `fake`. But it doesn't matter

tests/test_functional_post_install.py
```diff
@@ -127,6 +144,17 @@ def test_rose_fileinstall_uses_suite_defines(tmp_path):
             {
                 'test/rose-suite.conf': 'opts=a',
                 'test/opt/rose-suite-cylc-install.conf': '',
+                'test/opt/rose-suite-a.conf': '',
+                'test/opt/rose-suite-b.conf': '',
+                'test/opt/rose-suite-c.conf': '',
+                'ref/opt/rose-suite-cylc-install.conf': '!opts=b c\n',
+                'ref/rose-suite.conf': '!opts=a b c (cylc-install)',
```
MetRonnie, line 151: These two already in the dict from the previous commit?

tests/test_functional_post_install.py
```diff
@@ -108,10 +115,20 @@ def test_rose_fileinstall_uses_suite_defines(tmp_path):
...
                     '!opts=bar\n[env]\nBAR=1',
                 'ref/opt/rose-suite-cylc-install.conf':
                     '!opts=bar baz\n[env]\nBAR=2',
+                'ref/rose-suite.conf': '!opts=foo bar baz (cylc-install)',
+                'ref/opt/rose-suite-foo.conf': '',
+                'ref/opt/rose-suite-bar.conf': '',
+                'ref/opt/rose-suite-baz.conf': '',
+                'ref/log/18151210T0000Z-rose-suite.conf': (
+                    '!opts=bar baz\n\n[env]\nBAR=2\n'
+                ),
```
MetRonnie, line 131: I haven't understood why these are being added to the test

cylc/rose/utilities.py
```diff
@@ -442,3 +439,15 @@ def simplify_opts_strings(opts):
...
+    dumper = ConfigDumper()
+    timestamp = DateTimeOperator().process_time_point_str(
+        print_format='%Y%m%dT%H%M%S%Z'
+    )
+    fname = f'log/{timestamp}-rose-suite.conf'
```
MetRonnie, line 449: Shouldn't this be:
```suggestion
    fname = f'log/conf/{timestamp}-rose-suite.conf'
```
See https://github.com/cylc/cylc-rose/issues/10#issuecomment-762802718

cylc/rose/entry_points.py
wxtim: Yes. I've no idea why I've missed that.

tests/test_config_node.py
wxtim: Tidier.

tests/test_functional_post_install.py
wxtim: Ay

tests/test_functional_post_install.py
wxtim: The tests tests that the lists of files in the test and reference directory are the same: `assert titems == ritems` at around line 230.

cylc/rose/utilities.py
wxtim: Looks like it.

cylc/rose/utilities.py
```diff
@@ -442,3 +439,25 @@ def simplify_opts_strings(opts):
...
+        String filepath of the dump file relative to the install directory.
+    """
+    dumper = ConfigDumper()
+    timestamp = DateTimeOperator().process_time_point_str(
+        print_format='%Y%m%dT%H%M%S%Z'
+    )
+    fpath = dest_root / f'log/conf/{timestamp}-rose-suite.conf'
+    fpath.parent.mkdir(exist_ok=True, parents=True)
+    dumper.dump(node, str(fpath))
+    return f'log/conf/{timestamp}-rose-suite.conf'
```
MetRonnie, line 463: Sorry to be picky, but I would use a var here, otherwise if this gets broken in future, it might be a bit hard to debug
```suggestion
    rel_path = f'log/conf/{timestamp}-rose-suite.conf'
    fpath = dest_root / rel_path
    fpath.parent.mkdir(exist_ok=True, parents=True)
    dumper.dump(node, str(fpath))
    return rel_path
```

tests/test_functional_post_install.py
MetRonnie: Ok, was mainly wondering why so many

PR #38, opened 2021-02-23
AAA_Install rose suite.conf in cylc rose

### LINKED ISSUE #36: Install rose-suite.conf in cylc rose
**Describe exactly what you would like to see in an upcoming release**

Cylc-rose and _not_ Cylc-flow should install the ``rose-suite.conf`` file from the source directory rather than the work directory. This will allow more complex behavior from the plugin in future, such as diffing the old and new ``rose-suite.conf``

**Additional context**

You will need to add an exclusion in cylc-flow/cylc/flow/suite_files.py as a companion to changes to cylc-rose.

**Pull requests welcome!**

wxtim:
Closes #36, when run in combination with [cylc-flow PR 4094](https://github.com/cylc/cylc-flow/pull/4094)

### DIFF
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -20,6 +20,7 @@ Top level module providing entry point functions.
 """
 
 import os
+import shutil
 
 from pathlib import Path
 
@@ -182,6 +183,7 @@ def record_cylc_install_options(
     """
     if not rose_config_exists(dir_, opts):
         return False
+
     # Create a config based on command line options:
     cli_config = get_cli_opts_node(opts)
     # Leave now if there is nothing to do:
@@ -217,3 +219,42 @@ def record_cylc_install_options(
     )
     dumper(rose_suite_conf, rose_conf_filepath)
     return cli_config, rose_suite_conf
+
+
+def copy_config_file(
+    dir_=None,
+    opts=None,
+    dest_root=None,
+):
+    """Copy the ``rose-suite.conf`` from a workflow source to run directory.
+
+    Args:
+        dir_ (pathlib.Path | or str):
+            Source Path of Cylc install.
+        opts:
+            Not used in this function, but requried for consistent entry point.
+        dest_root (pathlib.Path | or str):
+            Destination path of Cylc install - the workflow rundir.
+
+    Return:
+        True if ``rose-suite.conf`` has been installed.
+        False if insufficiant information to install file given.
+    """
+    if (
+        dest_root is None or
+        dir_ is None
+    ):
+        return False
+
+    rundir = Path(dest_root)
+    srcdir = Path(dir_)
+    srcdir_rose_conf = srcdir / 'rose-suite.conf'
+    rundir_rose_conf = rundir / 'rose-suite.conf'
+
+    if not srcdir_rose_conf.is_file():
+        return False
+    elif rundir_rose_conf.is_file():
+        rundir_rose_conf.unlink()
+    shutil.copy2(srcdir_rose_conf, rundir_rose_conf)
+
+    return True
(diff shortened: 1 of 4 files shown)


tests/functional/test_copy_config_file.py
```diff
@@ -0,0 +1,87 @@
...
+            True
+        ),
+        (
+            # Valid sourcedir with rose file, rose file at dest:
+            {
+                'src/rose-suite.conf': '[env]\nFOO=2',
+                'dest/rose-suite.conf': '[env]\nFOO=1'
+            },
+            {'dir_': 'src', 'dest_root': 'dest'},
+            True
```
oliver-sanders, line 64: I think this is the only test we need here, just to make sure that the new contents are copied over the old.

oliver-sanders (approved):

tests/functional/test_copy_config_file.py
```diff
@@ -0,0 +1,87 @@
...
+        ),
+        (
+            # Valid sourcedir with rose file, no rose file at dest:
+            {
+                'src/rose-suite.conf': '[env]\nFOO=1',
+                'dest/jelly.txt': 'any-old-stuff'
+            },
+            {'dir_': 'src', 'dest_root': 'dest'},
+            True
+        ),
```
wxtim, line 56:
```suggestion
```

tests/functional/test_copy_config_file.py
wxtim: Killed the rest.

cylc/rose/entry_points.py
```diff
@@ -217,3 +219,28 @@ def record_cylc_install_options(
     )
     dumper(rose_suite_conf, rose_conf_filepath)
     return cli_config, rose_suite_conf
+
+
+def copy_config_file(
+    dest_root=None,
+    opts=None,
+    dir_=None,
+):
```
MetRonnie, line 228: Docstring 😋

cylc/rose/entry_points.py
```diff
@@ -217,3 +219,28 @@ def record_cylc_install_options(
...
+        dir_ is None
+    ):
+        return False
+
+    rundir = Path(dest_root)
+    srcdir = Path(dir_)
+    srcdir_rose_conf = srcdir / 'rose-suite.conf'
+    rundir_rose_conf = rundir / 'rose-suite.conf'
+
+    if not srcdir_rose_conf.is_file():
```
MetRonnie, line 240: One thing to bear in mind, `.is_file()` returns `True` if it's a symlink. I guess it's probably pretty unlikely that someone would make `rose-suite.conf` a symlink, but possible?

wxtim: If it is then we end up copying the target using `shutil.copy2(srcdir_rose_conf, rundir_rose_conf)`, which seems to me like reasonable behaviour.

cylc/rose/entry_points.py
```diff
@@ -226,6 +226,20 @@ def copy_config_file(
...
 ):
+    """Copy the ``rose-suite.conf`` from a workflow source to run directory.
+
+    Args:
+        dir_ (pathlib.Path | or str):
+            Source Path of Cylc install.
+        opts:
+            Not used in this function, but requried for consistent entry point.            
+        dest_root (pathlib.Path | or str):
+            Destination path of Cylc install - the workflow rundir.
```
MetRonnie, line 237: Thanks, but order is backwards

cylc/rose/entry_points.py
```diff
@@ -217,3 +219,42 @@ def record_cylc_install_options(
     )
     dumper(rose_suite_conf, rose_conf_filepath)
     return cli_config, rose_suite_conf
+
+
+def copy_config_file(
+    dest_root=None,
+    opts=None,
+    dir_=None,
+):
```
wxtim, line 228:
```suggestion
def copy_config_file(
    dir_=None,
    opts=None,
    dest_root=None,
):
```

cylc/rose/entry_points.py
wxtim: Fixed, but by changing the order of the kwargs. Bear in mind that I'll be refactoring this early next week. I think that the order should be "sourcedir", "opts", "rundir".

MetRonnie (approved):

PR #40, opened 2021-02-24
Cylc.rose pre release refactor

wxtim:
A generic tidy up of Cylc Rose, attempting to comment the code better and remove the odd unnecessary line.
Should not contain any functional changes.

Has [accompanying cylc-flow PR](https://github.com/cylc/cylc-flow/pull/4097)

- variable name change: dir_ => srcdir
- variable name change: dest_root => destdir
- Entry points to have just one top level function each.
- Top level functions ordered by order in which they are called.
- Move checks for input srcdir and destdir being filepaths or None
  to top level entry point functions to reduce duplication.
- Remove support for inputs to top level functions being str
- Tested path checking
- Re-read edited docstrings.
- Tidied up utils.py
- Removed rose-dump-log function from fileinstall to top level of the
  entry point becasue it's unrelated to fileinstall.

PR #42, opened 2021-02-26
Cylc reinstall  `--clear-rose-install-options`

### LINKED ISSUE #41: CLI option to clear previous rose install opts
### Description

See [rose cylc install proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-cylc-rose-installing-rose-configs.md#3-the-cylc-reinstall-command-should-have-an-option-to-delete-the-cylc-install-opt-conf): Cylc install should be able to clear existing configs saved in `opt/rose-suite-cylc-install.conf` thus:
```
cylc reinstall --clear-rose-install-options
```

### Additional context
This will require a PR in:
- in `cylc-rose` to add the required functionality.
- in `cylc-flow` to add the command line option.

wxtim:
Depends on [cylc-flow PR 4094](https://github.com/cylc/cylc-flow/pull/4094); tests _will_ fail with cylc-flow@master.

Closes #41 : Allow users to re-install a workflow without merging new CLI config with previous `opt/rose-suite-cylc-install.conf`.

Functional test equivalent to [scenario 3 in the proposal](https://github.com/cylc/cylc-admin/blob/master/docs/proposal-cylc-rose-installing-rose-configs.md#3-the-cylc-reinstall-command-should-have-an-option-to-delete-the-cylc-install-opt-conf).


tests/functional/test_reinstall_clean.py
```diff
@@ -0,0 +1,183 @@
...
+    monkeymodule.delenv('ROSE_SUITE_OPT_CONF_KEYS', raising=False)
+    result = subprocess.run(
+        [
+            'cylc', 'reinstall',
+            f'{fixture_provide_flow["test_flow_name"]}/run1',
+            '-O', 'baz', '-D', '[env]BAR=2'
+        ],
+        capture_output=True,
+        env=os.environ
+    )
```
MetRonnie, line 116: Looks to me like this is duplication of `test_reinstall.py`, unless I'm missing something? Why not just do an install and reinstall with `--clear-rose-install-options`?

tests/functional/test_reinstall.py
```diff
@@ -0,0 +1,204 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Functional tests for reinstalling of config files.
+This test does the following:
+
+1. Validates a workflow.
+2. Installs a workflow with some opts set using -O and
+   ROSE_SUITE_OPT_CONF_KEYS.
+3. Re-install workflow.
+4. After modifying the source ``rose-suite.conf``, re-install the flow again.
```
MetRonnie, line 23: Would be be useful to have these as comments/docstrings at the beginning of each test

wxtim: Why not: I'll do that.

tests/functional/test_reinstall_clean.py
wxtim: Yes, on consideration I agree.

MetRonnie (approved):

datamel:
I'm having trouble getting tests to pass locally with this branch, despite having the recommended cylc-flow (now merged https://github.com/cylc/cylc-flow/pull/4094) checked out. I have also tried with the https://github.com/cylc/cylc-flow/pull/4101 checked out with no success. For info I have pip installed etc.

MetRonnie:
Good point, I also get a lot of test failures, including `test_reinstall.py`

cylc/rose/entry_points.py
```diff
@@ -222,3 +228,44 @@ def record_cylc_install_options(
...
+
+    Return:
+        True if ``rose-suite.conf`` has been installed.
+        False if insufficiant information to install file given.
+    """
+    if (
+        dest_root is None or
+        dir_ is None
+    ):
+        raise FileNotFoundError(
```
wxtim, line 256: Before this returned False, which was causing failures in tests which didn't create both source and run directories. I think that this is safer.

cylc/rose/entry_points.py
```diff
@@ -222,3 +228,44 @@ def record_cylc_install_options(
     )
     dumper(rose_suite_conf, rose_conf_filepath)
     return cli_config, rose_suite_conf
+
+
+def copy_config_file(
+    dir_=None,
```
datamel, line 234: Worth changing `_dir` to `srcdir` and `dest_root` to `destdir` for consistency with : https://github.com/cylc/cylc-rose/pull/40

datamel (approved):

cylc/rose/entry_points.py
datamel: Tests passing locally for me now. Thanks Tim.

cylc/rose/entry_points.py
wxtim: No - these have to be the same for all the entry point functions and require a change in Cylc-flow to match, so doing this here would break things. 
I will be rebasing #40 once everything else is in.

PR #44, opened 2021-03-05
Fix failure of non-rose workflows

wxtim:
sibling: https://github.com/cylc/cylc-flow/pull/4097

Cause: Cylc rose trying to dump config where config == False.
Fix: Prevent cylc-rose dumping config if config not present.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,0 +1,29 @@
+# THIS FILE IS PART OF THE ROSE-CYLC PLUGIN FOR THE CYLC SUITE ENGINE.
+# Copyright (C) NIWA & British Crown (Met Office) & Contributors.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""Functional tests for top-level function record_cylc_install_options and
+"""
+
+import pytest
+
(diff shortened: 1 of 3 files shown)


oliver-sanders:
I think the other entry points should also have

```python
if not rose_config_exists(srcdir, opts):
```

clauses to prevent them from being run when the `rose-suite.conf` file isn't present.

wxtim:
>  I think the other entry points should also have `if not rose_config_exists(srcdir, opts):`

They already did.

oliver-sanders:
`def post_install` doesn't. I think if it did this error would be impossible?

wxtim:
> `def post_install` doesn't. I think if it did this error would be impossible?

Sorry, yes, all the next level down functions which used to be entry_points have it. _ Given that it's a common requirement it makes more sense in the entry point functions.

tests/conftest.py
```diff
@@ -0,0 +1,29 @@
...
+import pytest
+
+from cylc.flow import __version__ as CYLC_VERSION
+
+
+@pytest.fixture(scope='package', autouse=True)
+def set_cylc_version():
+    from _pytest.monkeypatch import MonkeyPatch
+    mpatch = MonkeyPatch()
+    yield mpatch.setenv('CYLC_VERSION', CYLC_VERSION)
```
datamel, line 28: Just curious if you know what was making it source the wrong cylc? Is this something users could run into?

datamel (approved):
I have checked this branch out with the sibling https://github.com/cylc/cylc-flow/pull/4097. I have read through the code, and run tests. I can confirm that this fixes broken tests, such as `tests/f/jobscript/00-torture.t ` and `tests/f/deprecations/03-suiterc.t` that are currently broken on cylc-flow master.

oliver-sanders:
Tests still failing on GH actions, tested myself locally and replicated the failures with the following:

cylc-flow master@07c7f5c9c
cylc-rose tim/pre-release-refactor@995ff8f
rose master@3afff091

oliver-sanders:
Needs this cylc-flow branch - https://github.com/cylc/cylc-flow/pull/4097

Tests aok.

oliver-sanders (approved):

oliver-sanders:
(will merge-squash)

PR #49, opened 2021-03-25
setup: load __version__ from package

oliver-sanders:
* Define the version number in the package.
  * This means we can do `cylc.rose.__version__`.
* Load this version number in the `setup.py` file.
  * This way we never get mismatching version numbers.
* Change the version from "1.0.0" to "0.1.0".
  * Use version "0" for code that is not production ready yet.
  * We will release version `1.0.0` to accompany Cylc `8.0.0` and Rose `2.0.0`.

```python
>>> import cylc.rose
>>> cylc.rose.__version__
'0.1.0'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ EXTRAS_REQUIRE['all'] = list(
 
 setup(
     name='cylc-rose',
-    version=__version__,
+    version=__version__,   # noqa
     long_description=long_description,
     long_description_content_type="text/markdown",
     install_requires=INSTALL_REQUIRES,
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -42,10 +45,9 @@
     }
 ) + TESTS_REQUIRE
 
-
 setup(
     name='cylc-rose',
-    version='1.0.0',
+    version=__version__,
```
wxtim, line 50: The obvious answer is probably to change to pycodestyle which seems to pass this, even though flake8 doesn't.

wxtim (changes requested):
Please can you fix the flake 8 issue?

wxtim (approved):

PR #56, opened 2021-04-23
suite -> workflow

MetRonnie:
Companion of https://github.com/cylc/cylc-flow/pull/4174, https://github.com/cylc/cylc-flow/pull/4198

### DIFF
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -1,4 +1,4 @@
-# THIS FILE IS PART OF THE ROSE-CYLC PLUGIN FOR THE CYLC SUITE ENGINE.
+# THIS FILE IS PART OF THE ROSE-CYLC PLUGIN FOR THE CYLC WORKFLOW ENGINE.
 # Copyright (C) NIWA & British Crown (Met Office) & Contributors.
 #
 # This program is free software: you can redistribute it and/or modify
@@ -146,11 +146,11 @@ def record_cylc_install_options(
             Cylc option parser object - we want to extract the following
             values:
             - opt_conf_keys (list or str):
-                Equivelent of ``rose suite-run --option KEY``
+                Equivalent of ``rose suite-run --option KEY``
             - defines (list of str):
-                Equivelent of ``rose suite-run --define KEY=VAL``
-            - suite_defines (list of str):
-                Equivelent of ``rose suite-run --define-suite KEY=VAL``
+                Equivalent of ``rose suite-run --define KEY=VAL``
+            - rose_template_var (list of str):
+                Equivalent of ``rose suite-run --define-suite KEY=VAL``
         rundir (pathlib.Path):
             Path to dump the rose-suite-cylc-conf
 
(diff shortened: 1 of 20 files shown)


cylc/rose/entry_points.py
```diff
@@ -149,8 +149,8 @@ def record_cylc_install_options(
                 Equivelent of ``rose suite-run --option KEY``
             - defines (list of str):
                 Equivelent of ``rose suite-run --define KEY=VAL``
-            - suite_defines (list of str):
-                Equivelent of ``rose suite-run --define-suite KEY=VAL``
+            - workflow_defines (list of str):
+                Equivelent of ``rose suite-run --define-workflow KEY=VAL``
```
wxtim, line 153: This is talking about a Rose 1/Cylc 7 command, so it needs to stay "suite".

tests/test_config_tree.py
```diff
@@ -42,27 +42,27 @@
 
 def test_rose_config_exists_no_dir(tmp_path):
     assert rose_config_exists(None, SimpleNamespace(
-        opt_conf_keys=None, defines=[], define_suites=[])
+        opt_conf_keys=None, defines=[], define_workflows=[])
```
wxtim, line 45: If you are going to change all of these I'd prefer `workflow_defines` - the original reflected `rose1` syntax, but I dislike it.

wxtim (commented):
I think that we need to seek clarity on Rose 2 nomenclature. I was under the impression that we were keeping the term "suite" WRT Rose. I don't actually like this choice.

cylc/rose/entry_points.py
wxtim, line 153:
```suggestion
Equivalent of ``rose suite-run --define-workflow KEY=VAL``
```
My spelling error.

cylc/rose/utilities.py
```diff
@@ -283,13 +283,13 @@ def get_cli_opts_node(opts=None):
...
-    suite_defines = []
+    rose_template_var = []
     if opts and 'opt_conf_keys' in dir(opts):
         opt_conf_keys = opts.opt_conf_keys
     if opts and 'defines' in dir(opts):
         defines = opts.defines
-    if opts and 'define_suites' in dir(opts):
-        suite_defines = opts.define_suites
+    if opts and 'rose_template_var' in dir(opts):
+        rose_template_var = opts.rose_template_var
```
wxtim, line 292: Sorry to be horribly pedantic - can we make this plural?

wxtim (approved):
1 non breaking grammatical comment. No Sig issue. 
Merge with or w'out responding to grammar comment.

cylc/rose/utilities.py
MetRonnie: Makes sense

MetRonnie: https://github.com/cylc/cylc-flow/pull/4200

datamel (approved):
Looks great, thanks @MetRonnie.
A couple of possible rogue `suites` remain
- https://github.com/cylc/cylc-rose/blob/77292ca865f3fc8ee828d348783d0ce658380fbb/cylc/rose/__init__.py#L71
- https://github.com/cylc/cylc-rose/blob/77292ca865f3fc8ee828d348783d0ce658380fbb/cylc/rose/stem.py#L5
- https://github.com/cylc/cylc-rose/blob/77292ca865f3fc8ee828d348783d0ce658380fbb/tests/functional/test_rose_stem.py#L5
- (I understand we have rose stem suites so I am assuming that the remaining suites in `stem.py`, e.g. check_suite_version, should remain as suite. Can you confirm @wxtim?)

MetRonnie:
> `# This file is part of Rose, a framework for meteorological suites.`

I think that should be left as is, because a forecasting suite is still a thing (I'm guessing it's what Cylc workflows were originally named after?)

PR #57, opened 2021-04-26
Fix rose orig host

wxtim:
This PR fixes an issue discussed verbally with @dpmatthews : the standard `ROSE_ORIG_HOST` variable is not available to Cylc Play. To make it available it has to be added to `opt/rose-suite-cylc-install.conf`.

To do this I needed to reconsider the templating engine detected logic in cylc-rose - We need to add a template variables section if none exists to put `ROSE_ORIG_HOST` in. We don't want to add it to `[jinja2:suite.rc]`.

This made me aware of a secondary issue: It was potentially possible to have an installed config file with multiple, incompatible templating engine sections:
```
[jinja2:suite.rc]
[empy:suite.rc]
```

*Manual test failing for @dpmatthews :*
Run a workflow on a remote host (such as our testing host) where:
```cylc
#!jinja2
...
[runtime]
  [[mytask]]
    platform = {{ROSE_ORIG_HOST}}
```
To setup the scheduler host:
```cylc
# global.cylc
[scheduler]
    [[run hosts]]
        available = els055
```

Check that task is being run on the box you started the workflow on, by checking jobfile or otherwise.


wxtim:
Failing on MyPy Checks fixed by [Cylc-Flow PR4196](https://github.com/cylc/cylc-flow/pull/4196)

tests/functional/test_play.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+@pytest.fixture(scope='module')
+def fixture_install_flow(fixture_provide_flow, monkeymodule):
+    """Run ``cylc install``.
+
+    By running in a fixture with modular scope we
+    can run tests on different aspects of its output as separate tests.
+
+    If a test fails using ``pytest --pdb then``
```
MetRonnie, line 115: I'm guessing this is meant to be
```suggestion
    If a test fails then using ``pytest --pdb`` and
```

tests/functional/13_cylc_play/suite.rc
```diff
@@ -0,0 +1,27 @@
+#!jinja2
```
MetRonnie, line 1: Shouldn't this file be named `flow.cylc`?

tests/functional/test_play.py
```diff
@@ -0,0 +1,190 @@
...
+        [
+            'cylc', 'install',
+            '--flow-name', fixture_provide_flow['test_flow_name'],
+            '-C', str(fixture_provide_flow['srcpath'])
+        ],
+        capture_output=True,
+        env=os.environ
+    )
+    yield {
+        'fixture_provide_flow': fixture_provide_flow,
```
MetRonnie, line 128: I think this would be easier to use if the `fixture_provide_flow` dict was expanded into this one
```suggestion
        **fixture_provide_flow,
```
(Obviously this change would need corresponding ones for the tests that use the fixture)

tests/functional/test_play.py
```diff
@@ -0,0 +1,190 @@
...
+
+@pytest.fixture(scope='module')
+def fixture_play_flow(fixture_install_flow):
+    """Run cylc flow in a fixture.
+    """
+    flowname = fixture_install_flow['fixture_provide_flow']['test_flow_name']
+    flowname = f"{flowname}/runN"
+    play = subprocess.run(
+        ['cylc', 'play', flowname, '--no-detach'],
+        capture_output=True
```
MetRonnie, line 141:
```suggestion
        capture_output=True, text=True
```
would eliminate the need for a `.decode()` later

cylc/rose/utilities.py
```diff
@@ -314,16 +334,21 @@ def get_cli_opts_node(opts=None):
...
+        config_node = rose_config_tree_loader(srcdir, opts).node
+        templating = identify_templating_section(config_node)
+    else:
+        templating = 'template variables'
+
     for define in rose_template_vars:
-        # For now just assuming that we just support Jinja2 - after I've
-        # Implemented the fully template-engine neutral template variables
-        # section this should be a moot point.
+        # TODO test what happens if CLI settings use "wrong" templating engine.
```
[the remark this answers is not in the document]
wxtim: A;ready dun
```suggestion
```

tests/functional/test_play.py
wxtim: Excellent Idea.

tests/functional/13_cylc_play/suite.rc
wxtim: 🐆

MetRonnie (approved):

tests/functional/test_reinstall.py
```diff
@@ -231,10 +239,44 @@ def test_cylc_reinstall_run2(fixture_reinstall_flow2):
...
+def test_cylc_reinstall_fail_on_clashing_template_vars(tmp_path):
+    """If you re-install with a different templating engine in suite.rc
+    reinstall should fail.
+    """
+    (tmp_path / 'rose-suite.conf').write_text(
+        '[jinja2:suite.rc]\n'
+        'Primrose=\'Primula Vulgaris\'\n'
+    )
+    (tmp_path / 'flow.cylc').touch()
+    test_flow_name = f'cylc-rose-test-{str(uuid4())[:8]}'
```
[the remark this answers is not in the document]
wxtim: I think that I have now fixed this.

wxtim:
>     * mix of jinja2 and empy resulted in error and traceback which didn't add to the diagnosis, perhaps this could just be a descriptive error.

Where did you mix jinja2 and empy?  I tried doing it in the `rose-suite.conf` and it was fine, I thought.

> traceback occurs if running a workflow that includes rose_orig_host but has not been previously installed, optional to change this to more user friendly error. 

This is true of any Jinja2 Variable which is not set. @oliver-sanders, should we have a special warning for the this one variable not being set? 
I'd hope that people who use ROSE_ORIG_HOST will always use Cylc Install as part of their work practices.

oliver-sanders:
> This is true of any Jinja2 Variable which is not set.

Yep, there's nothing more we can do.

It should still be possible to run Cylc commands on the source workflow (e.g. `validate`, `config`) due to the pre-configure plugin?

wxtim:
### Code behaviour

So, I've had to make some changes:
As the PR was we set the ROSE_ORIG_HOST variable in the pre-configure entry point. This had the advantage of making ROSE_ORIG_HOST available to play, validate and install. However, when `cylc play` is re-invoked on a remote server, the value of ROSE_ORIG_HOST is over-written with the name of that server, which defeats the whole point of having it.

To fix this I moved the setting ROSE_ORIG_HOST to being done in the `post-install` entry point, where it is written into the `rose-suite-cylc-install.conf` file. This means that it remains the same for Cylc Play where ever Cylc Play is run.

But there is a side effect: If Jinja2 in a `flow.cylc` file uses `{{ROSE_ORIG_HOST}}` the following will fail:
- `cylc validate <srcdir>`  (users might reasonably do this).
- `cylc play <scrcdir>`  (users probably shouldn't do this if their suite contains Rose elements).

To solve this problem I added extra logic back into the `pre-configure` entry point, where `ROSE_ORIG_HOST` is calculated if it isn't in the config.

Thanks to @dpmatthews  and @oliver-sanders  who helped me work my head around this.

MetRonnie:
> But there is a side effect: If Jinja2 in a `flow.cylc` file uses `{{ROSE_ORIG_HOST}}` the following will fail:
> - `cylc play <scrcdir>` (users probably shouldn't do this if their suite contains Rose elements).

But you can't do that, you can only do `cylc play <workflow-name>`

dpmatthews:
All seems to be working OK in my simple tests

MetRonnie (approved):
Tested and everything looks good as far as I can tell, it does run on the host specified in `global.cylc`

MetRonnie:
Wait, I tried your more detailed method from Teams:

**`~/cylc-src/temp4/flow.cylc :`**
```ini
#jinja2
[scheduling]
    [[graph]]
        R1 = foo
[runtime]
    [[foo]]
        platform = {{ROSE_ORIG_HOST}}
```
**`~/cylc-src/temp4/rose-suite.conf :`**
```ini
# (empty)
```
Installed:
```
$ cylc install temp4 --no-run-name
```
Ran:
```
$ cylc play temp4 -n --host=els055
WARNING: els055 is provided for the processing of OFFICIAL information
...
INFO - Cold Start 1
...
ERROR - [jobs-submit cmd] (platform not defined)
	[jobs-submit ret_code] 1
	[jobs-submit err] No matching platform "{{ROSE_ORIG_HOST}}" found

**Comment by dpmatthews** [comment]
@MetRonnie -  Should be `#!jinja2`

**Comment by MetRonnie** [comment]
D'oh

Fixed, now I get
```
ERROR - [jobs-submit cmd] (platform not defined)
	[jobs-submit ret_code] 1
	[jobs-submit err] No matching platform "vldXXX.YYY.metoffice.gov.uk" found
```
So I'm guessing that's fine, based on:
> If you installed your workflow on `mypc42` and play your suite with `--host=deepthought` then `ROSE_ORIG_HOST` should be `mypc42`

In `flow.cylc.processed` I get
```ini
[runtime]
    [[foo]]
        platform = vldXXX.YYY.metoffice.gov.uk
```

dpmatthews:
Yes, that's fine - we just need to update our site global config

datamel (approved):
Detailed review manual tests conducted with @wxtim over teams. 
Code read and locally have tests passing. 
Thanks @wxtim!

MetRonnie (approved):

PR #58, opened 2021-05-05
suite_defines → rose_template_vars follow up.

wxtim:


wxtim:
Forget it, Ronnie got this.

PR #61, opened 2021-05-25
Fix sphinx reference

MetRonnie:
Attempt to fix this [cylc-doc nightly build error](https://github.com/cylc/cylc-doc/runs/2661115314?check_suite_focus=true):
```
Warning, treated as error: 
unknown directive or role name: rose:ref 
make: *** [Makefile:37: html] Error 2 
```
The build process worked locally for me before and after this change, so not sure what's going on. Possibly a backwards incompatible change in a new Sphinx version?

### DIFF
--- a/cylc/rose/__init__.py
+++ b/cylc/rose/__init__.py
@@ -27,7 +27,7 @@ Rose Config
 -----------
 
 A fuller description of
-:rose:ref:`rose suite config is available here<Rose Suites>`.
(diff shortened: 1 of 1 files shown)


oliver-sanders:
> not sure what's going on

The doc tests / deploy thinggy use strict builds set via `SPHINXOPTS`. We don't do this by default as it makes developing a real pain.

```
make SPHINXOPTS='-Wn' clean html
```

oliver-sanders (approved):

PR #63, opened 2021-06-14
check root-dir

datamel:
Use of `root-dir` in a rose-suite.conf raises a warning. 
This PR also fixes the flake8 issues that were causing master to fail, and fixes mypy failure too with adding `--install-types` to the GH Actions mypy command.

To test `cylc install` a workflow with a rose-suite.conf containing root-dir config. This logic existed before but was skipping e.g.
 ```
root-dir{share/cycle}=$TMPDIR

### DIFF
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -104,12 +104,7 @@ def get_rose_vars(srcdir=None, opts=None):
     # Load the raw config tree
     config_tree = rose_config_tree_loader(srcdir, opts)
     # Warn if root-dir set in config
-    old_rose_suite_run_config = (
-        config_tree.node.get_value(["rose-suite-run"]))
-    list_to_check = list(config_tree.node)
-    if old_rose_suite_run_config:
-        list_to_check.extend(list(old_rose_suite_run_config))
-    for string in list_to_check:
+    for string in list(config_tree.node):
         if 'root-dir' in string:
             LOG.warning(
                 'You have set "root-dir", which is not supported at '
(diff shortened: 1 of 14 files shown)


cylc/rose/entry_points.py
```diff
@@ -103,13 +103,20 @@ def get_rose_vars(srcdir=None, opts=None):
...
-
-    # Warn if root-dir set in config:
-    if 'root-dir' in config_tree.node:
-        LOG.warning(
-            'You have set "root-dir", which at Cylc 8 does nothing. '
-            'See Cylc Install documentation.'
-        )
+    # Warn if root-dir set in config
+    old_rose_suite_run_config = (
+        config_tree.node.get_value(["rose-suite-run"]))
```
oliver-sanders, line 108: Just to confuse you the `root-dir` setting goes in a `rose-suite-run` section in the site-user config but *not* in the suite config.

* http://metomi.github.io/rose/doc/html/api/configuration/site-and-user.html#rose:conf:rose.conf[rose-suite-run]root-dir
* http://metomi.github.io/rose/doc/html/api/configuration/suite.html#rose:conf:rose-suite.conf|root-dir

datamel: Ah doh. I'll change this. Want me to change this to check the site-user config or not check?

oliver-sanders: I think we can ignore the site-user config.

setup.py
```diff
@@ -33,7 +33,7 @@
 }
 TESTS_REQUIRE = [
     'coverage>=5.0.0',
-    'flake8',
+    'flake8-*',
```
oliver-sanders, line 36: I don't think you can glob pypi modules, try replacing this with:

```
    'flake8-broken-line>=0.3.0',
    'flake8-bugbear>=21.0.0',
    'flake8-builtins>=1.5.0',
    'flake8-comprehensions>=3.5.0',
    'flake8-debugger>=4.0.0',
    'flake8-mutable>=1.2.0',
```

oliver-sanders (approved):

PR #72, opened 2021-07-30
fix tests where Cylc Flow is missing a shebang

wxtim:
Should fix testing failures.

Mutually exclusive https://github.com/cylc/cylc-flow/pull/4337


oliver-sanders (approved):
Lets go with this for the release.

PR #77, opened 2021-08-16
disabled capturing of stderr to allow us to see cause of failure in functional tests 

### LINKED ISSUE #71: Make Functional Tests Fail Helpfully
### Problem
Lots of tests take the form
```python
result = subprocess.run(['some', 'command'])
assert result.returncode == 0
```
If these tests fail you don't get a helpful error message.

### Possible Solutions:
- compare stderr, stdout and returncode tuple in test.
- log the stderr if test fails.

**Pull requests welcome!**

wxtim:
Closes #71 

Currently when functional test fails (often when it runs `cylc validate`) one has locate the pytest running that code and add breakpoints, even when the test failure relates to problems which are small or tangential to the feature in question.

By adding `-s` to the pytest option any failures will show the stdout from the commands run.

To test, try doing anything nasty (delete the runtime section/add an illegal config item/remove icp/change the name of a jinja2 template variable) to the `flow.cylc` in any of the functional test workflows and running the tests.

see also [Pytest docs: How to caputure stdout/err](https://pytest.org/en/latest/how-to/capture-stdout-stderr.html#setting-capturing-methods-or-disabling-capturing)

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -4,5 +4,7 @@ addopts = --verbose
     -n=1
     # group tests by module or class
     --dist=loadscope
+    # Disable stdout/err capturing so that we can see Cylc errors
+    -s
 testpaths =
     tests/


MetRonnie (approved):
This works ok, but another solution is to use `check=True` like so

```diff
--- a/tests/functional/test_pre_configure.py
+++ b/tests/functional/test_pre_configure.py
@@ -91,9 +91,7 @@ def test_validate(tmp_path, srcdir, envvars, args):
     script = ['cylc', 'validate', str(srcdir)]
     if args:
         script = script + args
-    assert (
-        run(script, env=envvars)
-    ).returncode == 0
+    run(script, env=envvars, check=True)
 
```

With this, pytest captures and prints the stderr after the traceback, instead of it showing up amongst the list of tests and their statuses.

> If check is true, and the process exits with a non-zero exit code, a `CalledProcessError` exception will be raised. Attributes of that exception hold the arguments, the exit code, and stdout and stderr if they were captured.

https://docs.python.org/3.7/library/subprocess.html#subprocess.run

wxtim:
> This works ok, but another solution is to use `check=True` like so
> 
> ```diff
> --- a/tests/functional/test_pre_configure.py
> +++ b/tests/functional/test_pre_configure.py
> @@ -91,9 +91,7 @@ def test_validate(tmp_path, srcdir, envvars, args):
>      script = ['cylc', 'validate', str(srcdir)]
>      if args:
>          script = script + args
> -    assert (
> -        run(script, env=envvars)
> -    ).returncode == 0
> +    run(script, env=envvars, check=True)
>  
> ```
> 
> With this, pytest captures and prints the stderr after the traceback, instead of it showing up amongst the list of tests and their statuses.
> 
> > If check is true, and the process exits with a non-zero exit code, a `CalledProcessError` exception will be raised. Attributes of that exception hold the arguments, the exit code, and stdout and stderr if they were captured.
> 
> https://docs.python.org/3.7/library/subprocess.html#subprocess.run

Preferable, but a much bigger job. I'll have a look at it this afternoon.

kinow (approved):

kinow:
Created follow-up issue: https://github.com/cylc/cylc-rose/issues/78 Thanks @wxtim

PR #86, opened 2021-10-07
Warn ROSE_VERSION in rose-suite.conf is ignored

wxtim:
After doing some checks for #11 

`ROSE_VERSION` is ignored if set in `rose-suite.conf`. This PR adds a warning message for users that this is the case.
`CYLC_VERSION` is ignored by Cylc, and already has a warning message.

### DIFF
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -92,7 +92,7 @@ def get_rose_vars_from_config_node(config, config_node, environ):
             LOG.warning(
                 f'[{section}]ROSE_VERSION={user_rose_version.value} '
                 'from rose-suite.conf will be ignored: '
-                f'ROSE_VERSION set: {ROSE_VERSION}'
+                f'Using Rose: {ROSE_VERSION}'
             )
         config_node[section].set(['ROSE_VERSION'], ROSE_VERSION)
 
@@ -100,7 +100,8 @@ def get_rose_vars_from_config_node(config, config_node, environ):
             user_cylc_version = config_node[section].value['CYLC_VERSION']
             LOG.warning(
                 f'[{section}]CYLC_VERSION={user_cylc_version.value} '
-                'from rose-suite.conf will be ignored.'
+                'from rose-suite.conf will be ignored: '
+                f'Using Cylc: {CYLC_VERSION}'
             )
         config_node[section].unset(['CYLC_VERSION'])
 
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -87,6 +87,13 @@ def get_rose_vars_from_config_node(config, config_node, environ):
     for section in ['env', templating]:
         # Add standard ROSE_VARIABLES
         config_node[section].set(['ROSE_SITE'], rose_site)
+        if 'ROSE_VERSION' in config_node[section].value:
+            user_rose_version = config_node[section].value['ROSE_VERSION']
+            LOG.warning(
+                f'ROSE_VERSION {user_rose_version.value} '
+                'from rose-suite.conf will be ignored: '
+                f'ROSE_VERSION set: {ROSE_VERSION}'
+            )
         config_node[section].set(['ROSE_VERSION'], ROSE_VERSION)
         if rose_orig_host:
             config_node[section].set(['ROSE_ORIG_HOST'], rose_orig_host)
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -24,7 +24,7 @@ import shlex
 from typing import TYPE_CHECKING, Union
 
 from cylc.flow.hostuserutil import get_host
-from cylc.flow import LOG
+from cylc.flow import LOG, __version__ as CYLC_VERSION
 from cylc.rose.jinja2_parser import Parser
 from metomi.rose import __version__ as ROSE_VERSION
 from metomi.isodatetime.datetimeoper import DateTimeOperator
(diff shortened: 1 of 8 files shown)


cylc/rose/utilities.py
```diff
@@ -87,6 +87,13 @@ def get_rose_vars_from_config_node(config, config_node, environ):
     for section in ['env', templating]:
         # Add standard ROSE_VARIABLES
         config_node[section].set(['ROSE_SITE'], rose_site)
+        if 'ROSE_VERSION' in config_node[section].value:
+            user_rose_version = config_node[section].value['ROSE_VERSION']
+            LOG.warning(
+                f'ROSE_VERSION {user_rose_version.value} '
```
oliver-sanders, line 93:
```suggestion
                f'[{section}]ROSE_VERSION={user_rose_version.value} '
```

cylc/rose/utilities.py
```diff
@@ -87,6 +87,13 @@ def get_rose_vars_from_config_node(config, config_node, environ):
     for section in ['env', templating]:
         # Add standard ROSE_VARIABLES
         config_node[section].set(['ROSE_SITE'], rose_site)
+        if 'ROSE_VERSION' in config_node[section].value:
+            user_rose_version = config_node[section].value['ROSE_VERSION']
```
oliver-sanders, line 91: Do we need to remove this variable from the config in order to prevent it from being exported?

cylc/rose/utilities.py
```diff
@@ -87,6 +87,13 @@ def get_rose_vars_from_config_node(config, config_node, environ):
     for section in ['env', templating]:
         # Add standard ROSE_VARIABLES
         config_node[section].set(['ROSE_SITE'], rose_site)
+        if 'ROSE_VERSION' in config_node[section].value:
```
oliver-sanders, line 90: Need to do the same for CYLC_VERSION.

cylc/rose/utilities.py
wxtim: We replace it in the config with `metomi.rose.__version__` which _is_ exported. Do we want this or not?

datamel (approved):
Thanks @wxtim. I have read the code, checked out the branch, run the tests and manually tested the change. 
image
One question: Do we want to inform users of the version of Cylc that will be used or how the cylc version is sourced for their info? 
No problems found for me.
If I had my picky hat on 'CYLC_VERSION' etc could be defined at the top for easy renaming in future. But that is non-essential and could be dealt with some other time as a wider PR to cover other string literals in cylc rose.

wxtim:
>  Do we want to inform users of the version of Cylc that will be used or how the cylc version is sourced for their info?

Certainly not here, since we only unset it.

cylc/rose/utilities.py
```diff
@@ -87,7 +87,23 @@ def get_rose_vars_from_config_node(config, config_node, environ):
...
+            user_rose_version = config_node[section].value['ROSE_VERSION']
+            LOG.warning(
+                f'[{section}]ROSE_VERSION={user_rose_version.value} '
+                'from rose-suite.conf will be ignored: '
+                f'ROSE_VERSION set: {ROSE_VERSION}'
```
oliver-sanders, line 95: I think this is a bit easier to understand:

```suggestion
                f'Using Rose: {ROSE_VERSION}'
```

cylc/rose/utilities.py
```diff
@@ -87,7 +87,23 @@ def get_rose_vars_from_config_node(config, config_node, environ):
...
+            )
         config_node[section].set(['ROSE_VERSION'], ROSE_VERSION)
+
+        if 'CYLC_VERSION' in config_node[section].value:
+            user_cylc_version = config_node[section].value['CYLC_VERSION']
+            LOG.warning(
+                f'[{section}]CYLC_VERSION={user_cylc_version.value} '
+                'from rose-suite.conf will be ignored.'
```
oliver-sanders, line 103: If we are reporting the Rose version we might as well report the Cylc version too:

```suggestion
                'from rose-suite.conf will be ignored: '
                f'Using Cylc: {CYLC_VERSION}'
```

Just need to add an import at the top:

```diff
 from cylc.flow.hostuserutil import get_host
-from cylc.flow import LOG
+from cylc.flow import (
+    LOG,
+    __version__ as CYLC_VERSION,
+)
 from cylc.rose.jinja2_parser import Parser
```

tests/test_config_node.py
```diff
@@ -67,6 +68,17 @@ def test_get_rose_vars_from_config_node__unbound_env_var(caplog):
     assert exc.match('env=X: MYVAR: unbound variable')
 
 
+def test_get_rose_vars_from_config_node__ignores_ROSE_VERSION(caplog):
+    """It should fail if variable unset in environment.
+    """
+    ret = {}
+    node = ConfigNode()
+    node.set(['template variables', 'ROSE_VERSION'], 99)
+    get_rose_vars_from_config_node(ret, node, {})
```
oliver-sanders, line 77: Can we add a test to ensure the `ROSE_VERSION` is being ovewritten correctly. Oh and can we test Cylc too.

cylc/rose/utilities.py
oliver-sanders: Oh, just spotted:

> Certainly not here, since we only unset it.

From the perspective of Cylc Rose that makes sense.

From the perspective of the user, they asked for "a" but got "b" so it might be a bit confusing not to report "b" in this case?

Although they will find out soon enough when they run their workflow...

tests/test_config_node.py
wxtim: Always up for more testing.

wxtim: Hopefully https://github.com/cylc/cylc-rose/pull/86/commits/150af3743c4b6da83ab6fafd67d971a3ad3bf666 answers...

wxtim:
@oliver-sanders - Were you yo be happy with this please squash-merge.

oliver-sanders (approved):

PR #87, opened 2021-10-13
Small MyPy fix

wxtim:
No related issue. Spotted when working on https://github.com/metomi/rose/pull/2485

Removed the Optional from return value of get_platform_from functions.
These functions should either fail with an error from the Cylc `get_platform`
function or return something. The output is **not** optional.

To test run:

```
cd rose
mypy --explicit-package-bases metomi/rose/suite_engine_procs/cylc.py
```

### DIFF
--- a/cylc/rose/platform_utils.py
+++ b/cylc/rose/platform_utils.py
@@ -19,7 +19,7 @@
 """
 from optparse import Values
 from pathlib import Path
-from typing import Optional, Dict, Any
+from typing import Dict, Any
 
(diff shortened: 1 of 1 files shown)


MetRonnie:
But `cylc.flow.platforms.get_platform()` has an `Optional` return?
https://github.com/cylc/cylc-flow/blob/256c585b1c6de8afd4ebf10afa20eb45ee2d8818/cylc/flow/platforms.py#L101-L105

PR #90, opened 2021-11-03
changed are name 'flow_name' to 'workflow_name' in line with changed …

wxtim:
…name in cylc-flow

This is a small change fixing a bug caused by a change of internal name for a CLI option: Because the name had changed Cylc Rose wasn't able to properly interact with it.

Tiny change ⇒ Single review ticket.

### DIFF
--- a/cylc/rose/stem.py
+++ b/cylc/rose/stem.py
@@ -447,7 +447,7 @@ class StemRunner:
             self.reporter(SuiteSelectionEvent(thissuite))
 
         # Create a default name for the suite; allow override by user
-        if not self.opts.flow_name:
+        if not self.opts.workflow_name:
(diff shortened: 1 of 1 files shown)


cylc/rose/stem.py
```diff
@@ -447,7 +447,7 @@ def process(self):
             self.reporter(SuiteSelectionEvent(thissuite))
 
         # Create a default name for the suite; allow override by user
-        if not self.opts.flow_name:
+        if not self.opts.workflow_name:
             self.opts.flow_name = self._generate_name()
```
wxtim, line 451:
```suggestion
            self.opts.workflow_name = self._generate_name()
```

MetRonnie (approved):

PR #91, opened 2021-11-09
setup: pin cylc-flow version

oliver-sanders:
* See https://github.com/cylc/cylc-admin/issues/130

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ with open("README.md", "r") as fh:
 
 INSTALL_REQUIRES = [
     'metomi-rose>=2.0b2',
-    'cylc-flow>=8.0b2',
+    'cylc-flow==8.0b3.*',  # the .* permits dev versions
 ]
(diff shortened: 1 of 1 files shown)


oliver-sanders:
Note, metomi-rose does not presently use the `.dev` convention.

MetRonnie (commented):
Do we not want to pin the metomi-rose version?

setup.py
```diff
@@ -28,7 +28,7 @@
 
 INSTALL_REQUIRES = [
     'metomi-rose>=2.0b2',
```
oliver-sanders, line 30:
```suggestion
    'metomi-rose==2.0b2',
```

oliver-sanders:
No reason, but will have to re do this again at release time anyway.

MetRonnie (approved):

MetRonnie:
Oh I see, because it's going to be 2.0b3?

PR #92, opened 2021-11-10
cylc-rose run twice bug

wxtim:
check whether workflow is a rose suite now can take CLI options to support workflow being a suite if rose CLI vars passed

Paired with https://github.com/cylc/cylc-flow/pull/4502

### DIFF
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -188,12 +188,13 @@ def rose_config_exists(
         return False
 
     # If _any_ of the following are true we want to return True.
+    # n.b. When the rose_config_tree_loader is called it
+    # will fail if `rose-suite.conf` doesn't exist.
     if (
-        Path(srcdir, 'rose-suite.conf').is_file() or (
-            getattr(opts, 'opt_conf_keys', False) or
-            getattr(opts, 'defines', False) or
-            getattr(opts, 'rose_template_vars', False)
-        )
+        Path(srcdir, 'rose-suite.conf').is_file() or
+        getattr(opts, 'opt_conf_keys', False) or
+        getattr(opts, 'defines', False) or
+        getattr(opts, 'rose_template_vars', False)
     ):
         return True
     return False
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -189,10 +189,11 @@ def rose_config_exists(
 
     # If _any_ of the following are true we want to return True.
     if (
-        Path(srcdir, 'rose-suite.conf').is_file() or
-        opts and opts.opt_conf_keys or
-        opts and opts.defines or
-        opts and opts.rose_template_vars
+        Path(srcdir, 'rose-suite.conf').is_file() or (
+            getattr(opts, 'opt_conf_keys', False) or
+            getattr(opts, 'defines', False) or
+            getattr(opts, 'rose_template_vars', False)
+        )
     ):
         return True
     return False
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -187,17 +187,8 @@ def rose_config_exists(
     if srcdir is None:
         return False
 
-    # If _any_ of the following are true we want to return True.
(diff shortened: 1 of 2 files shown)


cylc/rose/utilities.py
```diff
@@ -189,10 +189,18 @@ def rose_config_exists(
...
+            opts and hasattr(opts, 'opt_conf_keys')
+            and opts.opt_conf_keys
+        )
+        or (
+            opts and hasattr(opts, 'defines')
+            and opts.defines
+        ) or (
+            opts and hasattr(opts, 'rose_template_vars')
+            and opts.rose_template_vars
+        )
```
oliver-sanders, line 203: Can you explain this bit. If there's no `rose-suite.conf` file then there is nothing for these options to apply to?

wxtim: It should probably be if there is a rose suite.conf and ...

What do you think should happen if someone uses the rose options (-D and -S, -O is useless) to set template variables without there being a rose-suite.conf present? Error? Why shouldn't you be able to set rows variables on the command line only if you want?

cylc/rose/utilities.py
```diff
@@ -189,10 +189,18 @@ def rose_config_exists(
 
     # If _any_ of the following are true we want to return True.
     if (
-        Path(srcdir, 'rose-suite.conf').is_file() or
-        opts and opts.opt_conf_keys or
-        opts and opts.defines or
-        opts and opts.rose_template_vars
+        Path(srcdir, 'rose-suite.conf').is_file()
+        or (
```
wxtim, line 193:
```suggestion
        (
```

cylc/rose/utilities.py
```diff
@@ -189,10 +189,16 @@ def rose_config_exists(
 
     # If _any_ of the following are true we want to return True.
     if (
-        Path(srcdir, 'rose-suite.conf').is_file() or
-        opts and opts.opt_conf_keys or
-        opts and opts.defines or
-        opts and opts.rose_template_vars
+        (
+            opts and hasattr(opts, 'opt_conf_keys')
+            and opts.opt_conf_keys
```
MetRonnie, line 194: These can be simplifed to
```suggestion
            getattr(opts, 'opt_conf_keys', False)
```

cylc/rose/utilities.py
MetRonnie: I'm not sure you have resolved Oliver's question

cylc/rose/utilities.py
```diff
@@ -189,10 +189,15 @@ def rose_config_exists(
...
+        (
+            getattr(opts, 'opt_conf_keys', False)
+        ) or (
+            opts and hasattr(opts, 'defines')
+            and opts.defines
+        ) or (
+            opts and hasattr(opts, 'rose_template_vars')
+            and opts.rose_template_vars
+        )
     ):
```
MetRonnie, line 201: This block can still be tidied up - but I'm not sure you've resolved Oliver's point abut it: https://github.com/cylc/cylc-rose/pull/92#discussion_r746726875

cylc/rose/utilities.py
wxtim: @oliver-sanders  - From the context I think that removing the check immediately above this block for `rose-suite.conf` (which _is_ happening in Rose anyway so that condition was never used) answers your query - or have I missed something.

@MetRonnie  - Are you suggesting that the check for `rose-suite.conf` should remain and be corrected?

oliver-sanders: Sorry, I'm still confused.

I don't see how the `cylc install` options are relevant here, should this be:

```python
def rose_config_exists(srcdir: Union[Path, str, None], opts: 'Values'):
    return Path(srcdir, 'rose-suite.conf').is_file()
```

MetRonnie (approved):
Tests passing locally (install step failing on GH Actions due to unrelated reasons)

oliver-sanders (approved):

oliver-sanders:
(rerunning tests)

PR #94, opened 2021-11-11
setup: move project metadata to setup.cfg

oliver-sanders:
* Move metadata from setup.py to setup.cfg for simplicity.
* Keep the setup.py file, when we are ready we can remove this to switch to PEP517 build.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,7 +21,7 @@ long_description = file: README.md
 long_description_content_type = text/markdown
 author = British Crown (Met Office) & Contributors
 author_email = metomi@metoffice.gov.uk
-url=https://cylc.org/cylc-doc/latest/html/plugins/cylc-rose.html
+url=https://cylc.github.io/cylc-doc/latest/html/plugins/cylc-rose.html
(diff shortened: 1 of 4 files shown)


setup.cfg
```diff
@@ -16,9 +16,12 @@
 
 [metadata]
 name = cylc-rose
+version = attr: cylc.rose.__version__
+long_description = file: README.md
+long_description_content_type = text/markdown
 author = British Crown (Met Office) & Contributors
 author_email = metomi@metoffice.gov.uk
-url=https://cylc.github.io/cylc-doc/latest/html/plugins/cylc-rose.html
+url=https://cylc.org/cylc-doc/latest/html/plugins/cylc-rose.html
```
wxtim, line 24: Have you tried this link?

wxtim (commented):
I'm not sure that the URL works the way you think it works... Otherwise fine.

oliver-sanders:
Looks like we have a problem with the `cylc.org` redirection.

wxtim (approved):

PR #97, opened 2021-11-24
install rose in editable mode to avoid symlink issues

wxtim:


wxtim:
Broken becuause of Rose-stem changes

PR #100, opened 2021-12-06
GH Actions: Edit test schedule

MetRonnie:
I keep getting notifications whenever the tests fail on the scheduled run. This is because I merged the last PR that touched this line. If you merge this PR normally (i.e. not squash/rebase merge) then you should get the notifications from now on.


wxtim (approved):
Change legit: Unsure about cause of failure, will investigate elsewhere.

PR #104, opened 2022-01-07
Retire rose site

wxtim:
Closes #13

PR #106, opened 2022-01-12
Small re-write of special variables docs.

wxtim:
Follow up #104 
- [x] Fix spelling mistakes (i before e except in their).
- [x] Re-write.

Ended up with massive mission creep:

### Summary of discussion with @dpmatthews  and @oliver-sanders 

Cylc Rose should do the following with ROSE_ORIG_HOST:
- Annotate the config item in `rose-suite-cylc-install.conf` to indicate that the install process saved this config.
- Use the saved ROSE_ORIG_HOST for installed workflows.
- Make ROSE_ORIG_HOST the hostname of the machine running the script otherwise, warning the user that any value which they may have set will be over-ridden.

### DIFF
--- a/cylc/rose/__init__.py
+++ b/cylc/rose/__init__.py
@@ -99,7 +99,7 @@ to the Cylc scheduler:
       ``ROSE_VERSION`` in your suite configuration it will be ignored.
 
 
-.. caution::
+.. deprecated:: 8.0.0
 
    ``CYLC_VERSION`` will be removed from your configuration by the
    Cylc-Rose plugin, as it is now set by Cylc.
--- a/cylc/rose/__init__.py
+++ b/cylc/rose/__init__.py
@@ -77,7 +77,7 @@ The Cylc Rose plugin provides two environment/template variables
 to the Cylc scheduler:
 
 ``ROSE_ORIG_HOST``
-   Cylc scripts (such as ``cylc validate`` and ``cylc play``)
+   Cylc commands (such as ``cylc validate`` and ``cylc play``)
    will provide the name of the host on which the script is run,
    unless the workflow has been installed using ``cylc install``.
    If this is the case the value set in
@@ -85,12 +85,12 @@ to the Cylc scheduler:
 
    .. caution::
 
-      Therefore, Cylc commands run on uninstalled workflows may produce
+      Therefore, running Cylc commands on non-installed workflows may produce
       inconsistent values for ``ROSE_ORIG_HOST``.
 
 
 ``ROSE_VERSION``
-   When running Cylc scripts such as ``cylc install``,
+   When running Cylc commands such as ``cylc install``,
    ``cylc play`` and ``cylc validate``
    the plugin provides the version number of your installed Rose Version in
    workflow scheduler's environment.
--- a/cylc/rose/__init__.py
+++ b/cylc/rose/__init__.py
@@ -79,8 +79,9 @@ to the Cylc scheduler:
 ``ROSE_ORIG_HOST``
    Cylc scripts (such as ``cylc validate`` and ``cylc play``)
    will provide the name of the host on which the script is run,
-   unless the workflow has been installed using ``cylc install``. If this is
-   the case the value set in ``opt/rose-suite-cylc-install.conf`` will be used.
+   unless the workflow has been installed using ``cylc install``.
+   If this is the case the value set in
+   ``opt/rose-suite-cylc-install.conf`` will be used.
 
    .. caution::
 
@@ -89,7 +90,8 @@ to the Cylc scheduler:
 
 
 ``ROSE_VERSION``
-   When running Cylc scripts such as `cylc install`, `cylc play` and `cylc validate`
+   When running Cylc scripts such as ``cylc install``,
+   ``cylc play`` and ``cylc validate``
    the plugin provides the version number of your installed Rose Version in
    workflow scheduler's environment.
 
--- a/cylc/rose/__init__.py
+++ b/cylc/rose/__init__.py
@@ -77,16 +77,16 @@ The Cylc Rose plugin provides two environment/template variables
 to the Cylc scheduler:
 
 ``ROSE_ORIG_HOST``
-   Cylc commands (such as ``cylc validate`` and ``cylc play``)
-   will provide the name of the host on which the script is run,
-   unless the workflow has been installed using ``cylc install``.
-   If this is the case the value set in
-   ``opt/rose-suite-cylc-install.conf`` will be used.
+   Cylc commands (such as ``cylc install``, ``cylc validate`` and ``cylc play``)
+   will provide the name of the host on which the command is run.
 
-   .. caution::
+   If the workflow is installed the value of ``ROSE_ORIG_HOST`` will be
+   set in ``opt/rose-suite-cylc-install.conf`` and used by future commands
+   e.g. ``cylc play``.
 
-      Therefore, running Cylc commands on non-installed workflows may produce
-      inconsistent values for ``ROSE_ORIG_HOST``.
+   Using Cylc install should produce a more consistent value for
+   ``ROSE_ORIG_HOST``; running Cylc commands on uninstalled
+   workflows may produce inconsistent values.
 
 
 ``ROSE_VERSION``
--- a/cylc/rose/__init__.py
+++ b/cylc/rose/__init__.py
@@ -88,7 +88,9 @@ to the Cylc scheduler:
 
 
 ``ROSE_VERSION``
-   The plugin provides the version number of your installed Rose Version.
+   When running Cylc scripts such as `cylc install`, `cylc play` and `cylc validate`
+   the plugin provides the version number of your installed Rose Version in
+   workflow scheduler's environment.
 
    .. deprecated:: 8.0.0
 
--- a/cylc/rose/__init__.py
+++ b/cylc/rose/__init__.py
@@ -77,7 +77,8 @@ The Cylc Rose plugin provides two environment/template variables
 to the Cylc scheduler:
 
 ``ROSE_ORIG_HOST``
-   Cylc scripts will provide the name of the host on which the script is run,
+   Cylc scripts (such as ``cylc validate`` and ``cylc play``)
+   will provide the name of the host on which the script is run,
    unless the workflow has been installed using ``cylc install``. If this is
    the case the value set in ``opt/rose-suite-cylc-install.conf`` will be used.
 
--- a/cylc/rose/__init__.py
+++ b/cylc/rose/__init__.py
@@ -73,39 +73,28 @@ permitted in the ``rose-suite.conf`` files:
 Special Variables
 -----------------
 
-The Cylc Rose plugin has specific logic for handling a small group of special
-environment / template variables in the ``rose-suite.conf`` file:
+The Cylc Rose plugin provides two environment/template variables
+to the Cylc scheduler:
 
 ``ROSE_ORIG_HOST``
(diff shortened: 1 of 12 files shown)


cylc/rose/__init__.py
```diff
@@ -73,39 +73,28 @@
...
 
-The Cylc Rose plugin has specific logic for handling a small group of special
-environment / template variables in the ``rose-suite.conf`` file:
+The Cylc Rose plugin provides two environment/template variables
+to the Cylc scheduler:
 
 ``ROSE_ORIG_HOST``
-   The plugin provides the hostname of the computer where the plugin runs as
-   an evironment variable.
+   The plugin provides the hostname of the computer where the plugin runs.
```
oliver-sanders, line 80: Would be clearer to relate this back to the command the user ran. Can't remember if that's the `cylc install` or `cylc play` command though...

```
   The plugin provides the hostname of the computer where the plugin runs. I.E. the one where `cylc install` is run.
```

oliver-sanders (approved):

cylc/rose/__init__.py
wxtim: Both.

wxtim: @oliver-sanders  Have a look at my lastest update

cylc/rose/__init__.py
```diff
@@ -77,7 +77,19 @@
 to the Cylc scheduler:
 
 ``ROSE_ORIG_HOST``
-   The plugin provides the hostname of the computer where the plugin runs.
+   Cylc scripts will provide this as the name of the host the script is run
+   on, unless it is set in the ``[env]`` section of the configuration file.
```
oliver-sanders, line 81: > unless it is set in the ``[env]`` section

What!

wxtim: I just went and checked bc @dpmatthews asked me to.

cylc/rose/__init__.py
```diff
@@ -73,39 +73,28 @@
...
-   by Cylc's config processing.
-
-   .. deprecated:: 8.0.0
-
-      Setting ``[env]CYLC_VERSION`` in ``rose-suite.conf``.
-
-      With Cylc 7 / Rose2019 users could set ``CYLC_VERSION`` for thier
-      suites. This is no longer possible, and if you set ``CYLC_VERSION``
-      in your suite configuration will be overwritten.
+.. caution::
```
oliver-sanders, line 102: Can make this a deprecated item the same as for `ROSE_VERSION`.

cylc/rose/__init__.py
oliver-sanders: I think this would be clearer if we relate this back to the command they ran (because they don't know where or when the plugin runs, the "plugin" isn't really a user-facing concept in the way that `cylc install` and `cylc play` are).

cylc/rose/__init__.py
wxtim:
```suggestion
.. deprecated:: 8.0.0
```

cylc/rose/__init__.py
```diff
@@ -73,39 +73,36 @@
...
+   .. caution::
+
+      Therefore, Cylc commands run on uninstalled workflows may produce
+      inconsistent values for ``ROSE_ORIG_HOST``.
+
 
 ``ROSE_VERSION``
-   The plugin provides ``ROSE_VERSION`` from your installed Rose Version
-   in the environment section and any templating sections you have defined.
+   The plugin provides the version number of your installed Rose Version.
```
wxtim, line 91:
```suggestion
   When running Cylc scripts such as `cylc install`, `cylc play` and `cylc validate`
   the plugin provides the version number of your installed Rose Version in
   workflow scheduler's environment.
```

cylc/rose/__init__.py
```diff
@@ -73,39 +73,36 @@
...
 
-The Cylc Rose plugin has specific logic for handling a small group of special
-environment / template variables in the ``rose-suite.conf`` file:
+The Cylc Rose plugin provides two environment/template variables
+to the Cylc scheduler:
 
 ``ROSE_ORIG_HOST``
-   The plugin provides the hostname of the computer where the plugin runs as
-   an evironment variable.
+   Cylc scripts will provide the name of the host on which the script is run,
```
wxtim, line 80:
```suggestion
   Cylc scripts (such as ``cylc validate`` and ``cylc play``)
   will provide the name of the host on which the script is run,
```

cylc/rose/utilities.py
```diff
@@ -523,3 +530,81 @@ def paths_to_pathlib(paths):
...
+        >>> override_this_variable(node, 'env', 'ROSE_ORIG_HOST')
+        False
+    """
+    if variable not in node[section]:
+        return False
+    elif (
+        variable != 'ROSE_ORIG_HOST'
+        or (
+            variable in node[section]
+            and ROSE_ORIG_HOST_INSTALLED_OVERRIDE_STRING not in
```
MetRonnie, line 573: Think that's extraneous as we're already inside the `if variable not in node[section]: ... elif:`
```suggestion
            ROSE_ORIG_HOST_INSTALLED_OVERRIDE_STRING not in
```

cylc/rose/__init__.py
```diff
@@ -73,39 +73,41 @@
...
+   ``cylc play`` and ``cylc validate``
+   the plugin provides the version number of your installed Rose Version in
+   workflow scheduler's environment.
 
    .. deprecated:: 8.0.0
 
       Setting ``[env]ROSE_VERSION`` in ``rose-suite.conf``.
 
-      With Cylc 7 / Rose2019 users could set ``ROSE_VERSION`` for thier
+      With Cylc 7 / Rose2019 users could set ``ROSE_VERSION`` for their
```
wxtim, line 102:
```suggestion
      Setting ``[env]ROSE_VERSION`` in ``rose-suite.conf``.
      With Cylc 7 / Rose2019 users could set ``ROSE_VERSION`` for their
```

wxtim, line 102:
```suggestion
      With Cylc 7 / Rose2019 users could set ``ROSE_VERSION`` for their
```

tests/functional/test_reinstall.py
```diff
@@ -124,15 +126,24 @@ def test_cylc_install_run(fixture_install_flow):
...
     ]
 )
 def test_cylc_install_files(fixture_install_flow, file_, expect):
     fpath = fixture_install_flow['fixture_provide_flow']['flowpath']
-    assert (fpath / file_).read_text() == expect
+    result_text = re.sub(
+        'ROSE_ORIG_HOST=.*',
+        'ROSE_ORIG_HOST=<HOSTNAME>',
+        (fpath / file_).read_text()
+    )
```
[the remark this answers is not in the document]
wxtim: Yes. I was just getting hacked off with the GHA failures. :(

Will put it back as was .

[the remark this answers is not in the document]
wxtim: I think you are right, but the weird hostname issues on  GH Actions means that these tests are only working locally. :(

cylc/rose/__init__.py
```diff
@@ -73,39 +73,41 @@
...
+to the Cylc scheduler:
 
 ``ROSE_ORIG_HOST``
-   The plugin provides the hostname of the computer where the plugin runs as
-   an evironment variable.
+   Cylc commands (such as ``cylc validate`` and ``cylc play``)
+   will provide the name of the host on which the script is run,
+   unless the workflow has been installed using ``cylc install``.
+   If this is the case the value set in
+   ``opt/rose-suite-cylc-install.conf`` will be used.
```
oliver-sanders, line 84: This makes it sound like `cylc install` does *not* set `ROSE_ORIG_HOST`.

```suggestion
   Cylc commands (such as ``cylc install``, ``cylc validate`` and ``cylc play``)
   will provide the name of the host on which the command is run.

   If the workflow is installed the value of ``ROSE_ORIG_HOST`` will be
   set in ``opt/rose-suite-cylc-install.conf`` and used by future commands
   e.g. ``cylc play``.
```

cylc/rose/__init__.py
```diff
@@ -73,39 +73,41 @@
...
+
+   .. caution::
+
+      Therefore, running Cylc commands on non-installed workflows may produce
+      inconsistent values for ``ROSE_ORIG_HOST``.
```
oliver-sanders, line 89: The dangling "therefore" is confusing. Ideally combine with the preceding sentence (it's not important that this info is in a special box).

wxtim, line 89:
```suggestion
   Using Cylc install should produce a more consistent value for
   ``ROSE_ORIG_HOST``; running Cylc commands on uninstalled
   workflows may produce inconsistent values.
```

MetRonnie:
What is being done about the fact that the failing tests expect a partial domain name of the host but the installed workflow has FQDN?
```
E           [env]
E         - ROSE_ORIG_HOST=fv-az292-438.internal.cloudapp.net
E         + ROSE_ORIG_HOST=fv-az292-438.tqgtyn10aqxuxoi0mkfbldi5vh.jx.internal.cloudapp.net
```

MetRonnie:
Could a workaround be to only compare the first bit of the hostname (everything up to the first `.`)?

oliver-sanders:
So long as the cylc-rose code and related tests are both using the same hostname interface (provided by cylc-flow) then we are doing all that we can.

If we are getting different answers in different circumstances it's a platform problem.

cylc/rose/__init__.py
```diff
@@ -73,39 +73,41 @@
...
-   The plugin provides the hostname of the computer where the plugin runs as
-   an evironment variable.
+   Cylc commands (such as ``cylc install``, ``cylc validate`` and ``cylc play``)
+   will provide the name of the host on which the command is run.
+
+   If the workflow is installed the value of ``ROSE_ORIG_HOST`` will be
+   set in ``opt/rose-suite-cylc-install.conf`` and used by future commands
+   e.g. ``cylc play``.
+
+   Using Cylc install should produce a more consistent value for
```
MetRonnie, line 87:
```suggestion
   Using ``cylc install`` should produce a more consistent value for
```

cylc/rose/__init__.py
```diff
@@ -73,39 +73,41 @@
...
+   ``ROSE_ORIG_HOST``; running Cylc commands on uninstalled
+   workflows may produce inconsistent values.
```
MetRonnie, line 89: Maybe a clearer explanation of why might it produce inconsistent values?

MetRonnie (commented):
```
./cylc/rose/__init__.py:80:80: E501 line too long (80 > 79 characters)
```

cylc/rose/__init__.py
```diff
@@ -77,16 +77,18 @@
...
 
    If the workflow is installed the value of ``ROSE_ORIG_HOST`` will be
    set in ``opt/rose-suite-cylc-install.conf`` and used by future commands
    e.g. ``cylc play``.
 
-   Using Cylc install should produce a more consistent value for
-   ``ROSE_ORIG_HOST``; running Cylc commands on uninstalled
-   workflows may produce inconsistent values.
+   Using ``cylc install`` should produce a more consistent value
+   for ``ROSE_ORIG_HOST``; running Cylc commands on uninstalled
```
MetRonnie, line 89: Sorry for nitpicking, but I feel like "uninstalled" means it was... uninstalled, as opposed to a workflow that has not been *in*stalled at all
```suggestion
   for ``ROSE_ORIG_HOST``; running Cylc commands on non-installed
```

MetRonnie (approved):

oliver-sanders:
Two approvals, tested as working.

PR #109, opened 2022-01-24
Fix coverage; use flake8-simplify; update GH Actions

MetRonnie:
Need to do an editable pip install (`-e`) for coverage to work

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,10 +11,12 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


MetRonnie:
Coverage back up to 63% - https://app.codecov.io/gh/cylc/cylc-rose/compare/109/tree/cylc/rose (actually median more like 90%, `stem.py` is dragging down the mean)

oliver-sanders (approved):

MetRonnie:
Do we want to add Codecov as an app integration to this repo, like we have for cylc-flow?

PR #98, opened 2021-11-25
git clone rose and install editably

wxtim:

PR #114, opened 2022-02-17
bump version

oliver-sanders:


MetRonnie (approved):

MetRonnie (changes requested):
Also need to update?
https://github.com/cylc/cylc-rose/blob/7e7dfd3f98c9b9d3e602faa9e863cb1dd1eedac5/setup.cfg#L58-L59

oliver-sanders:
That would be an idea wouldn't it :facepalm:.

setup.cfg
```diff
@@ -55,8 +55,8 @@ packages = find_namespace:
 python_requires = >=3.7
 include_package_data = True
 install_requires =
-    metomi-rose==2.0rc1
-    cylc-flow==8.0rc1.*  # the .* permits dev versions
+    metomi-rose==2.0rc2
```
MetRonnie, line 58: Ah wait, looks like it was only the cylc-flow version that needed updating
```suggestion
    metomi-rose==2.0rc1
```

oliver-sanders: metomi-rose will need updating too, just hasn't been bumped yet.

MetRonnie (approved):

setup.cfg
oliver-sanders, line 58: I've shifted metomi-rose over to the `.dev` convention (especially as it was what caused me to jigger up the 1.0.0 release)

```suggestion
    metomi-rose==2.0rc2.*  # the .* permits dev versions
```

oliver-sanders:
Looks like a command failure in `tests/functional/test_pre_configure.py`, can't replicate locally.

.github/workflows/tests.yml
```diff
@@ -28,20 +28,15 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Install cylc-flow
+      - name: Install deps
         run: |
-          pip install 'cylc-flow[all] @ git+https://github.com/cylc/cylc-flow@master'
+          pip install 'cylc-flow @ git+https://github.com/cylc/cylc-flow@master'
+          pip install 'metomi-rose @ git+https://github.com/metomi/rose@master'
```
MetRonnie, line 34: Revert of https://github.com/cylc/cylc-rose/pull/98 ... but why?

oliver-sanders: It was ok at that PR but at some point after an additional `- name: Install Rose` step was added.

The `git clone` approach was used because of that symlink issue where `pip install <url>` was broken for Rose. Seems like it works fine now though so going back to the simpler approach.

.github/workflows/tests.yml
```diff
@@ -28,20 +28,15 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
-      - name: Install cylc-flow
+      - name: Install deps
         run: |
-          pip install 'cylc-flow[all] @ git+https://github.com/cylc/cylc-flow@master'
+          pip install 'cylc-flow @ git+https://github.com/cylc/cylc-flow@master'
```
MetRonnie, line 33: I think the tests are failing because empy is not getting installed

```suggestion
          pip install 'cylc-flow[all] @ git+https://github.com/cylc/cylc-flow@master'
```

MetRonnie (changes requested):

.github/workflows/tests.yml
MetRonnie: Either that or we add empy to the "tests" requirements for cylc-rose

oliver-sanders:
```suggestion
          pip install 'cylc-flow[empy] @ git+https://github.com/cylc/cylc-flow@master'
```

oliver-sanders: Lets give it a go.

MetRonnie (approved):

PR #115, opened 2022-02-18
Update release workflow

MetRonnie:
- Add tests badge [tests](https://github.com/cylc/cylc-rose/actions/workflows/tests.yml) to release PRs so we can be more aware of test failures before creating a release
- Update release date in changelog (depends on #112)

### DIFF
--- a/.github/workflows/1_create_release_pr.yml
+++ b/.github/workflows/1_create_release_pr.yml
@@ -12,6 +12,7 @@ on:
         default: 'master'
       check_shortlog:
         description: Check contributors list against the mailmap
+        type: boolean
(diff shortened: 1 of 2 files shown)


MetRonnie:
Ugh, ignore the test failures, they are irrelevant

wxtim:
> Ugh, ignore the test failures, they are irrelevant

Yes, I've seen these before - do we ken why they happen?

wxtim (approved):

MetRonnie:
In this case, it was the codecov uploader flaking out - https://github.com/codecov/codecov-action/issues/598

PR #120, opened 2022-03-16
Fcm make fix for remote and subshell platforms

datamel:
Sibling to: https://github.com/metomi/rose/pull/2557

Again credit to @wxtim for pair programming this.
These changes close https://github.com/metomi/rose/issues/2553

- [x] Applied any dependency changes to `setup.cfg` - None to apply.
- [x] No change log entry required (why? invisible to users).

### DIFF
--- a/cylc/rose/platform_utils.py
+++ b/cylc/rose/platform_utils.py
@@ -19,7 +19,7 @@
 """
 from optparse import Values
 from pathlib import Path
-from typing import Dict, Any
+from typing import Dict, Any, Union
 
 from cylc.flow.config import WorkflowConfig
 from cylc.flow.exceptions import PlatformLookupError
@@ -29,8 +29,8 @@ from cylc.flow.platforms import get_platform
 
 
 def get_platform_from_task_def(
-    flow: str, task: str
-) -> Dict[str, Any]:
+    flow: str, task: str, quiet: bool = False
+) -> Union[Dict[str, Any], str]:
     """Return the platform dictionary for a particular task.
 
     Uses the flow definition - designed to be used with tasks
@@ -47,12 +47,24 @@ def get_platform_from_task_def(
     config = WorkflowConfig(flow, flow_file, Values())
     # Get entire task spec to allow Cylc 7 platform from host guessing.
     task_spec = config.pcfg.get(['runtime', task])
-    platform = get_platform(task_spec)
+    try:
+        platform = get_platform(task_spec)
+    except PlatformLookupError:
+        # host is a subshell
+        host = task_spec['remote']['host']
+        if host is not None:
+            return host
+        else:
+            return 'localhost'
     if platform is None:
-        raise PlatformLookupError(
-            'Platform lookup failed; platform is a subshell to be evaluated: '
-            f' Task: {task}, platform: {task_spec["platform"]}.'
-        )
+        if not quiet:
+            raise PlatformLookupError(
+                'Platform lookup failed; platform is a subshell to be'
+                f' evaluated: Task: {task}, platform: {task_spec["platform"]}.'
+            )
+        else:
+            # platform is a subshell
+            return task_spec["platform"]
     return platform
 
 
--- a/tests/test_platform_utils.py
+++ b/tests/test_platform_utils.py
@@ -128,6 +128,9 @@ def fake_flow():
                     host = cheese
             [[kanga]]
                 platform = $(echo "myplatform")
+            [[roo]]
+                [[[remote]]]
+                    host = $(echo "myhost")
             [[BAR]]
                 platform = milk
             [[child_of_bar]]
@@ -195,6 +198,23 @@ def test_get_platform_from_task_def_raises(
     mock_glbl_cfg(*MOCK_GLBL_CFG)
     with pytest.raises(PlatformLookupError, match='Platform lookup failed.*'):
         get_platform_from_task_def(fake_flow[0], 'kanga')
+
+
+@pytest.mark.parametrize(
+    'task, expected',
+    [
+        pytest.param('kanga', '$(echo "myplatform")', id='platform subshell'),
+        pytest.param('roo', '$(echo "myhost")',  id='remote host subshell'),
+    ]
+)
+def test_get_platform_from_task_def_quiet(
+    mock_glbl_cfg, fake_flow, task, expected
+):
+    """Test getting platform from task definition with platform subshell.
+    """
+    mock_glbl_cfg(*MOCK_GLBL_CFG)
+    actual = get_platform_from_task_def(fake_flow[0], task, quiet=True)
+    assert actual == expected
 
 
 @pytest.mark.parametrize(


cylc/rose/platform_utils.py
```diff
@@ -47,12 +47,24 @@ def get_platform_from_task_def(
     config = WorkflowConfig(flow, flow_file, Values())
     # Get entire task spec to allow Cylc 7 platform from host guessing.
     task_spec = config.pcfg.get(['runtime', task])
-    platform = get_platform(task_spec)
+    try:
+        platform = get_platform(task_spec)
+    except PlatformLookupError:
+        # host is a subshell
```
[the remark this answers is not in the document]
wxtim: I think that the comment should make it cleared that the host being a subshell is only one possible cause of PlatformLookupError: The subsequent logic should probably check whether host is in the form of either `$(command)` or `` `command` ``.

I don't know why the platform logic checks for `platform = $(subshell)` but doesn't do the same for the host command, but I think making things more consistent might take up more time than is reasonable for functionality which is supposed to be deleted at Cylc 9 anyway.

cylc/rose/platform_utils.py
```diff
@@ -47,15 +49,30 @@ def get_platform_from_task_def(
...
     task_spec = config.pcfg.get(['runtime', task])
+    # check for subshell and evaluate
+    if 'platform' in task_spec and (
+        task_spec['platform'] is not None) and (
+            is_platform_definition_subshell(task_spec['platform'])):
+        task_spec['platform'] = eval_subshell(task_spec['platform'])
+    elif 'remote' in task_spec and (
+        'host' in task_spec['remote']) and (
+        task_spec['remote']['host'] is not None) and (
+            bool(HOST_REC_COMMAND.match(task_spec['remote']['host']))):
```
MetRonnie, line 60: Simplification
```suggestion
    elif (
        task_spec.get('remote', {}).get('host', None) is not None and 
        HOST_REC_COMMAND.match(task_spec['remote']['host'])
    ):
```

cylc/rose/platform_utils.py
```diff
@@ -47,15 +49,30 @@ def get_platform_from_task_def(
     config = WorkflowConfig(flow, flow_file, Values())
     # Get entire task spec to allow Cylc 7 platform from host guessing.
     task_spec = config.pcfg.get(['runtime', task])
+    # check for subshell and evaluate
+    if 'platform' in task_spec and (
+        task_spec['platform'] is not None) and (
+            is_platform_definition_subshell(task_spec['platform'])):
```
MetRonnie, line 55: Simplification
```suggestion
    if (
        task_spec.get('platform', None) is not None and
        is_platform_definition_subshell(task_spec['platform'])
    ):
```

cylc/rose/platform_utils.py
```diff
@@ -47,15 +49,30 @@ def get_platform_from_task_def(
...
     return platform
 
 
+def eval_subshell(platform):
+    """Evaluates platforms/hosts defined as subshell"""
+    match = HOST_REC_COMMAND.match(platform)
+    output = subprocess.run(
+        ['bash', '-c', f" {match[2]}"], capture_output=True
+    )
+    return output.stdout.strip().decode()
```
MetRonnie, line 73: Why the f-string? 

Also, `text=True` avoids the need to decode later
```suggestion
    output = subprocess.run(
        ['bash', '-c', match[2]], capture_output=True, text=True
    )
    return output.stdout.strip()
```

cylc/rose/platform_utils.py
```diff
@@ -17,20 +17,22 @@
...
 from optparse import Values
 from pathlib import Path
-from typing import Dict, Any
+from typing import Any, Dict
 
 from cylc.flow.config import WorkflowConfig
-from cylc.flow.exceptions import PlatformLookupError
 from cylc.flow.id_cli import parse_id
+from cylc.flow.platforms import (
+    HOST_REC_COMMAND,
```
MetRonnie, line 28: This allows backticks, was that desired?

oliver-sanders: This is only for the old `[remote]host` syntax which I think we supported at Cylc 7.

cylc/rose/platform_utils.py
```diff
@@ -47,15 +49,31 @@ def get_platform_from_task_def(
     config = WorkflowConfig(flow, flow_file, Values())
     # Get entire task spec to allow Cylc 7 platform from host guessing.
     task_spec = config.pcfg.get(['runtime', task])
+    # check for subshell and evaluate
+    if (
+        task_spec.get('platform', None) is not None and
+        is_platform_definition_subshell(task_spec['platform'])
```
oliver-sanders, line 55: In some cases parsec might give us an empty string rather than None. For example you get an empty string if you try to unset the platform like so:

```
[runtime]
    [[task]]
        platform =
```

So I think it should be safer to do `if platform_string` than `if platform_string is None`:

```suggestion
        task_spec.get('platform')
        and is_platform_definition_subshell(task_spec['platform'])
```

cylc/rose/platform_utils.py
```diff
@@ -47,15 +49,31 @@ def get_platform_from_task_def(
...
+    ):
+        task_spec['platform'] = eval_subshell(task_spec['platform'])
+    elif (
+        task_spec.get('remote', {}).get('host', None) is not None and 
+        HOST_REC_COMMAND.match(task_spec['remote']['host'])
```
oliver-sanders, line 60: Ditto:

```suggestion
        task_spec.get('remote', {}).get('host')
        and HOST_REC_COMMAND.match(task_spec['remote']['host'])
```

cylc/rose/platform_utils.py
```diff
@@ -47,15 +49,31 @@ def get_platform_from_task_def(
...
+    ):
+        task_spec['remote']['host'] = eval_subshell(
+            task_spec['remote']['host'])
     platform = get_platform(task_spec)
-    if platform is None:
-        raise PlatformLookupError(
-            'Platform lookup failed; platform is a subshell to be evaluated: '
-            f' Task: {task}, platform: {task_spec["platform"]}.'
-        )
```
oliver-sanders, line 55: What's the deal with removing this exception?

(Sorry if I've missed the discussion on this)

oliver-sanders (commented):
Looks good, going to try to get the test passing locally.

cylc/rose/platform_utils.py
datamel: This is no longer going to be hit. Before the return was None if platform was a subshell. Now, the subshell evaluation is made before checking for platform so will no longer return None.

MetRonnie (approved):
Happy, although I haven't tested. Is there an example workflow/Rose suite I can test manually? (The subshell bit, irrespective of FCM Make?)

wxtim:
> Happy, although I haven't tested. Is there an example workflow/Rose suite I can test manually? (The subshell bit, irrespective of FCM Make?)

t/rose-task-run/08-app-fcm-make/

cylc/rose/platform_utils.py
oliver-sanders: Right, ok, it only returns `None` if it *is* a subshell, ta.

oliver-sanders (approved):
Ran against the tests in metomi-rose, all good.

Issue #123, opened 2022-03-22
Deprecation warning for `[<engine>:suite.rc]`


wxtim:
**Describe the bug**
Setting `[jinja2:suite.rc]` in `rose-suite.conf` does not emit a deprecation warning.

**Steps to reproduce the bug**

Unzip 
[rose-dep-warnings.tar.gz](https://github.com/cylc/cylc-rose/files/8323659/rose-dep-warnings.tar.gz) into `~/cylc-src.

```
cylc validate rose-dep-warnings
Valid for cylc-8.0rc2.dev
```

**Expected behavior**
```
cylc validate rose-dep-warnings
Warning: [jinja2:suite.rc] is deprecated and will be removed: use [template variables] instead.
Valid for cylc-8.0rc2.dev
```

**Screenshots**

**Additional context**

It may be tidier to add an additional checking function to `ultilities.py` which can be called from `entry_points.py`.

To close this issue should also check:
- [ ] `[empy:suite.rc]`
- [ ] Passing both `[jinja2:suite.rc]` and `[empy:suite.rc]` to `-D` should also result in a warning.

**Pull requests welcome!**
This is an Open Source project - please consider contributing a bug fix
yourself (please read `CONTRIBUTING.md` before starting any work though).

PR #136, opened 2022-04-07
Log warning [jinja2/empy:suite.rc] is deprecated

datamel:
These changes close https://github.com/cylc/cylc-rose/issues/123

## Requirements check-list

- [x] No change log entry required (why? e.g. only a warning).

### DIFF
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -27,6 +27,7 @@ from pathlib import Path
 from metomi.rose.config import ConfigLoader, ConfigDumper
 from cylc.rose.utilities import (
     ROSE_ORIG_HOST_INSTALLED_OVERRIDE_STRING,
+    deprecation_warnings,
     dump_rose_log,
     get_rose_vars_from_config_node,
     identify_templating_section,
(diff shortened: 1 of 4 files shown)


wxtim (changes requested):
I've just tried this and I have a couple of things which I don't think are quite right:

#### Define section once, get warning twice 

```bash
> cat rose-suite.conf 
[jinja2:suite.rc]
> cylc install 
INSTALLED temp1/run10 from /net/home/h02/tpilling/cylc-src/temp1
2022-04-08T08:30:09+01:00 WARNING - [jinja2:suite.rc] is deprecated. Use [template variables] instead.
2022-04-08T08:30:09+01:00 WARNING - [jinja2:suite.rc] is deprecated. Use [template variables] instead.
```

#### It only works for Cylc install:

Same workflow and rose-suite.conf:

```bash
> cylc validate .
Valid for cylc-8.0rc3.dev
```

I think that you need to run the check in a function called by `cylc.rose.entry_points.pre_configure` such as `cylc.rose.entry_points.get_rose_vars` - around line 125 there is a check for a deprecated root_dir setting - I think here would be a reasonable place to put it - You might even create a new method `cylc.rose.utils.deprecation_warnings()` and move the existing warning into it...

wxtim (approved):
Thank you @datamel 
Merge if you wish - neither of my comments should block that!

datamel:
Feedback addressed. Thanks @wxtim.

datamel:
Feedback addressed. Thanks @wxtim.

PR #139, opened 2022-04-28
Fix rose stem

wxtim:
These changes addess the following significant issue:

### Issue

The `rose stem` CLI no longer paralleled the `cylc install` CLI and was breaking as a result.

### Proposed solution

Add a function converting CLI args to the appropriate opts for use by Cylc install.

## Requirements check-list

- [x] Appropriate tests are included (unit and/or functional) _and_ Already covered by existing tests.
- [x] I have updated the documentation in this PR branch.


MetRonnie:
```
import file mismatch:
imported module 'test_rose_stem' has this __file__ attribute:
  /home/runner/work/cylc-rose/cylc-rose/tests/test_rose_stem.py
which is not the same as the test file we want to collect:
  /home/runner/work/cylc-rose/cylc-rose/tests/functional/test_rose_stem.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```
Looks like the tests directly in `/tests/` need to be moved into `/tests/unit/`

MetRonnie:
> Looks like the tests directly in `/tests/` need to be moved into `/tests/unit/`

That wasn't it. Maybe an `__init__.py` in `/tests/` was what it wanted

tests/unit/test_rose_stem_units.py
```diff
@@ -0,0 +1,57 @@
...
+        sourcepath = (Path.cwd() / 'foo')
+    elif relpath:
+        sourcepath = (Path.cwd() / 'rose-stem')
+    else:
+        sourcepath = (Path.cwd() / 'foo')
+        args.append(str(sourcepath))
+    expect = SimpleNamespace(source=str(sourcepath))
+    sourcepath.mkdir()
+    result = get_source_opt_from_args(opts, args)
+    assert result == expect
```
MetRonnie, line 57: This looks like Too Much Logic In The Test Function to me

cylc/rose/stem.py
```diff
@@ -451,6 +452,23 @@ def process(self):
         return self.opts
 
 
+def get_source_opt_from_args(opts, args):
+    """Convert sourcedir given as arg or implied by no arg to opts.source.
+    """
+    if len(args) == 1:
+        # sourcedir not given:
+        opts.source = str(Path.cwd() / 'rose-stem')
```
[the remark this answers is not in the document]
wxtim: The assumption with rose stem (As far as I can tell) that you are installing a rose stem suite in a subdirectory of CWD.

tests/functional/test_rose_stem.py
```diff
@@ -200,17 +200,33 @@ def rose_stem_run_template(setup_stem_repo):
...
+            if verbosity > 2:
+                # If -vv print replication instructions and breakpoint.
+                msg = (
+                    'To reproduce failure outside test environment:'
+                    f'\n\tcd {setup_stem_repo["workingcopy"]}'
+                    f'\n\texport FCM_CONF_PATH={os.environ["FCM_CONF_PATH"]}'
+                    f'\n\t{rose_stem_cmd}'
+                )
+                print(msg)
+                breakpoint()  # noqa T100
```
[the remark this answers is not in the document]
wxtim:
```suggestion
# If you want to debug add a breakpoint here:
```

cylc/rose/stem.py
```diff
@@ -451,6 +452,23 @@ def process(self):
         return self.opts
 
 
+def get_source_opt_from_args(opts, args):
+    """Convert sourcedir given as arg or implied by no arg to opts.source.
```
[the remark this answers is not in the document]
wxtim: That's correct as far as I can tell...

Perhaps I should go quiz a power user/real code owner (we've been left holding someone else's baby here).

tests/unit/test_rose_stem_units.py
wxtim: I've tried to simplify this - _and I have got rid of the extra bool_ - but ultimately I don't think it's possible to specify everything I want in the parametrization bit - because the absolute path should probably point to the test dir.

datamel (approved):
Checked out and tested. I can't spot any problems with the implementation. Thanks @wxtim

cylc/rose/stem.py
```diff
@@ -20,10 +20,21 @@
...
+
+Looks for a suite to install either in $PWD/rose-stem, or in a specified
+location - e.g:
+
+- ``rose stem``:
+        Install a rose-stem suite from PWD/rose-stem
+- ``rose stem relative-path`` :
+        Install a rose-stem suite from PWD/relative-path.
+- ``rose stem /absoulte/path``:
+        Install a rose-stem suite from specified abs path.
```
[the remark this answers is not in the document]
MetRonnie: And would be good to have `rose stem [PATH]` as the first line of the docstring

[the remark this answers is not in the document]
wxtim: Done this.

wxtim:
@MetRonnie  - I think I've answered your points -

You are right about existing code having logic to handle `opts.source==False`, but not `hasattr(opts, 'source')==False`, so I've added that to the test: I think that is reasonable. 

Coming back to a new week I've managed to simplify the unit tests a good deal too. 😄

wxtim:
For the record, I've double checked with Steve Wardle, and if no suite source is specified we assume that the user wants `$PWD/rose-stem`

cylc/rose/stem.py
```diff
@@ -435,7 +450,7 @@ def process(self):
                         elements[0], '"' + elements[1] + '"')
 
         # Change into the suite directory
-        if self.opts.source:
+        if hasattr(self.opts, 'source') and self.opts.source:
```
MetRonnie, line 453:
```suggestion
        if getattr(self.opts, 'source', None):
```

cylc/rose/stem.py
```diff
@@ -451,6 +466,34 @@ def process(self):
...
+        Cylc options with source attribute added.
+    """
+    if len(args) == 0:
+        # sourcedir not given:
+        return opts
+    elif os.path.isabs(args[-1]):
+        # sourcedir given, and is abspath:
+        opts.source = args[-1]
+    else:
+        # breakpoint()
```
MetRonnie, line 490:
```suggestion
```

tests/unit/test_rose_stem_units.py
```diff
@@ -0,0 +1,59 @@
...
+    [
+        pytest.param(
+            [],
+            '',
+            id='no-path'
+        ),
+        pytest.param(
+            ['/foo'],
+            '/foo',
+            id='relative-path'
```
MetRonnie, line 36:
```suggestion
            id='absolute-path'
```

tests/unit/test_rose_stem_units.py
```diff
@@ -0,0 +1,59 @@
...
+        ),
+        pytest.param(
+            ['foo'],
+            '{tmp_path}/foo',
+            id='absolute-path'
```
MetRonnie, line 41:
```suggestion
            id='relative-path'
```

cylc/rose/stem.py
datamel: This is here to help future devs debug. There is a note in the code https://github.com/cylc/cylc-rose/blob/d669a2efb00e05f13efa544a7d8cb4d66a42adb3/tests/functional/test_rose_stem.py#L57

MetRonnie: That is referring to this actually
https://github.com/cylc/cylc-rose/blob/b0eb4a775927f904a4f604ee08240c09a4d51052/tests/functional/test_rose_stem.py#L229

MetRonnie (approved):
Thanks

datamel (approved):
Checked recent changes. Merging now with two approvals.

PR #140, opened 2022-04-29
Jinja2 v3

oliver-sanders:
> Sibling to an upcoming cylc-flow PR switching from Jinja2 v2->v3.

Jinja2v2 used to support zero-prefixed integers e.g. 01, 02, 03, ...

Jinja2v3 dropped support for this, fair enough, however, our users haven't been getting warnings for this so haven't had time to adapt.

This PR patches the Jinja2 parser we use in cylc-rose to shave off the leading zero before we pass the template variable back through to Jinja2 itself. It's not a particularly nice solution, it is intended as a stopgap to help users to migrate with the intention that we would retire it ASAP, probably with the move to Jinja2 3.1.

## Requirements check-list

- [x] Made Discourse post to raise awareness of this change - https://cylc.discourse.group/t/jinja2-upgrade-to-version-3/489
  (note post not yet public, waiting for the rc4 release first)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,7 @@
 
 [#140](https://github.com/cylc/cylc-rose/pull/140) -
 Support integers with leading zeros (e.g `001`) to back support Rose
-configurations for use with cylc-flow>=8.0rc5 which uses Jinja2 v3 which
+configurations for use with cylc-flow>=8.0rc4 which uses Jinja2 v3 which
 no longer supports this.
 
 ## __cylc-rose-1.0.3 (<span actions:bind='release-date'>Released 2022-05-20</span>)__
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,21 +1,30 @@
 # Selected Cylc-Rose Changes
 
+## __cylc-rose-1.? (<span actions:bind='release-date'>Pending</span>)__
+
+### Fixes
+
+[#140](https://github.com/cylc/cylc-rose/pull/140) -
+Support integers with leading zeros (e.g `001`) to back support Rose
+configurations for use with cylc-flow>=8.0rc5 which uses Jinja2 v3 which
+no longer supports this.
+
 ## __cylc-rose-1.0.3 (<span actions:bind='release-date'>Released 2022-05-20</span>)__
 
 ### Fixes
 
-[139](https://github.com/cylc/cylc-rose/pull/139) - Make `rose stem` command
+[#139](https://github.com/cylc/cylc-rose/pull/139) - Make `rose stem` command
 work correctly with changes made to `cylc install` in
 [cylc-flow PR #4823](https://github.com/cylc/cylc-flow/pull/4823)
 
-[130](https://github.com/cylc/cylc-rose/pull/130) - Fix bug preventing
+[#130](https://github.com/cylc/cylc-rose/pull/130) - Fix bug preventing
 ``cylc reinstall`` using Rose fileinstall.
 
-[132](https://github.com/cylc/cylc-rose/pull/132) - Fix bug preventing
+[#132](https://github.com/cylc/cylc-rose/pull/132) - Fix bug preventing
 Cylc commands (other than `install`) from accessing the content of
 `--rose-template-variable`.
 
-[133](https://github.com/cylc/cylc-rose/pull/133) - Fix bug allowing setting
+[#133](https://github.com/cylc/cylc-rose/pull/133) - Fix bug allowing setting
 multiple template variable sections.
 
 ## __cylc-rose-1.0.2 (<span actions:bind='release-date'>Released 2022-03-24</span>)__
@@ -25,7 +34,7 @@ multiple template variable sections.
 [118](https://github.com/cylc/cylc-rose/pull/118) - Fail if
 a workflow is not a Rose Suite but user provides Rose CLI options.
 
-## cylc-rose-1.0.1 (Released 2022-02-17)
+## __cylc-rose-1.0.1 (Released 2022-02-17)__
 
 First official release of Cylc-Rose.
 
--- a/cylc/rose/jinja2_parser.py
+++ b/cylc/rose/jinja2_parser.py
@@ -16,6 +16,8 @@
 """Utility for parsing Jinja2 expressions."""
 
 from ast import literal_eval as python_literal_eval
+from copy import deepcopy
+from contextlib import contextmanager
 import re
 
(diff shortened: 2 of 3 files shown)


oliver-sanders:
(note the tests will fail until cylc-flow moves to Jinja2 v3)

oliver-sanders:
ATM this PR supports a single leading zero e.g. 01, 02, 03,  ...

Looks like there is some usage of multiple leading zeros e.g. 001, 0002, 00003, ...

This diff should expand support to the latter cases:

```
diff --git a/cylc/rose/jinja2_parser.py b/cylc/rose/jinja2_parser.py
index 96cebae..9f95567 100644
--- a/cylc/rose/jinja2_parser.py
+++ b/cylc/rose/jinja2_parser.py
@@ -50,8 +50,11 @@ def _lexer_wrap(fcn):
                 and len(value_str) > 1
                 and value_str[0] == '0'
             ):
+                for ind, char in enumerate(value_str):
+                    if char != '0':
+                        break
                 instances.add(value_str)
-                value_str = value_str[1:]  # strip the leading zero
+                value_str = value_str[ind:]  # strip the leading zero
             yield (lineno, token, value_str)
 
     def _inner(
```

MetRonnie:
That diff does the trick 👍 

However, noticed the warning you get should be 
```diff
 WARNING - Support for integers with leading zeros was dropped in Jinja2 v3. Rose will extend support until a future version.
     Please amend your Rose configuration files e.g:
      * 0500 => 500
      * 050 => 50
      * 0900 => 900
-     * 0001 => 001
+     * 0001 => 1
      * 0100 => 100
```

oliver-sanders:
FYI tested zero patted integers with Jinja2 v2 under Python 2.7 and 3.7 to make sure they behaved the same which they did:

```console
$ cat suite.rc
#!Jinja2

{% set number = 007 %}

# {{ number }}
# {{ number + 2 }}
# {{ number + 002 }}

{% set number = 0099 %}

# {{ number }}
# {{ number + 2 }}
# {{ number + 002 }}

$ cylc view . --stdout --process
# 7
# 9
# 9
# 99
# 101
# 101
```

(no octal wierdness)

Jinja2 v3 octals take the same format as Python 3 octals.

cylc/rose/jinja2_parser.py
```diff
@@ -27,6 +29,155 @@
...
+    """
+    # clear any previously cashed lexer instances
+    jinja2.lexer._lexer_cache.clear()
+
+    # apply the code patch (new lexer instances will pick up these changes)
+    _integer_re = deepcopy(jinja2.lexer.integer_re)
+    jinja2.lexer.integer_re = re.compile(
+        rf'''
+            (
+                [0-9](_?\d)* # decimal
```
MetRonnie, line 143: This regex is a little confusing, may be worth commenting

```suggestion
                [0-9](_?\d)* # modified jinja2.lexer.integer_re to support leading 0s
```

cylc/rose/jinja2_parser.py
```diff
@@ -123,6 +277,8 @@ def literal_eval(self, value):
         # dump this value into a Jinja2 template
         templ = '{{ %s }}' % value
         # check that this expression consists only of literals
+        # with patch_jinja2_leading_zeros():
+        #     ast = self.parse(templ)
```
MetRonnie, line 281: Should this be removed?
```suggestion
```

cylc/rose/jinja2_parser.py
```diff
@@ -27,6 +29,155 @@
     Neg,
     Pos
 )
+import jinja2.lexer
+
+from cylc.flow import LOG
+
+
+def _strip_leading_zeros(string):
+    """Strip leading zeros from a string.
```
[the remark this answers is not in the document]
oliver-sanders: Nice.

CHANGES.md
```diff
@@ -1,21 +1,30 @@
 # Selected Cylc-Rose Changes
 
+## __cylc-rose-1.? (<span actions:bind='release-date'>Pending</span>)__
+
+### Fixes
+
+[#140](https://github.com/cylc/cylc-rose/pull/140) -
+Support integers with leading zeros (e.g `001`) to back support Rose
+configurations for use with cylc-flow>=8.0rc5 which uses Jinja2 v3 which
```
MetRonnie, line 9:
```suggestion
configurations for use with cylc-flow>=8.0rc4 which uses Jinja2 v3 which
```

oliver-sanders: Got ahead of myself there.

CHANGES.md
```diff
@@ -1,21 +1,30 @@
 # Selected Cylc-Rose Changes
 
+## __cylc-rose-1.? (<span actions:bind='release-date'>Pending</span>)__
+
+### Fixes
+
+[#140](https://github.com/cylc/cylc-rose/pull/140) -
+Support integers with leading zeros (e.g `001`) to back support Rose
+configurations for use with cylc-flow>=8.0rc4 which uses Jinja2 v3 which
+no longer supports this.
```
MetRonnie, line 10: Also this sentence is a bit difficult to parse, I would suggest something like
```suggestion
(For use with cylc-flow>=8.0rc4 which uses Jinja2 v3)
Maintain support for integers with leading zeros (e.g `001`)
in Rose configurations. Note this support is deprecated and you
should remove leading zeros.
```

MetRonnie (approved):

cylc/rose/jinja2_parser.py
```diff
@@ -27,6 +29,155 @@
     Neg,
     Pos
 )
+import jinja2.lexer
+
+from cylc.flow import LOG
+
+
+def _strip_leading_zeros(string):
```
wxtim, line 37: Is it reasonable to have a function which wraps another function with almost no change?
Is it reasonable to have 4 tests which ought to be covered by lstrip's unit testing?

oliver-sanders: * It's used twice so may as well centralise.
* The tests are more for documentation.

wxtim: Is there any risk of negative numbers being fed to this?

oliver-sanders: I think the sign will be stripped by this point but give it a try.

wxtim (approved):

cylc/rose/jinja2_parser.py
wxtim: I'm happy with this having tried it.

PR #142, opened 2022-05-23
Error if rose-suite-cylc-install.conf exists and not a re-install

wxtim:
These changes partially address https://github.com/cylc/cylc-flow/issues/4888

## Requirements check-list

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,14 @@
 # Selected Cylc-Rose Changes
 
+## __cylc-rose-1.0.4 (<span actions:bind='release-date'></span>)__
+
+### Fixes
+
+[#142](https://github.com/cylc/cylc-rose/pull/142) - Raise exception if
+cylc install finds an existing `rose-suite-cylc-install.conf` during an
+installation.
+
+
 ## __cylc-rose-1.0.3 (<span actions:bind='release-date'>Released 2022-05-20</span>)__
 
 ### Fixes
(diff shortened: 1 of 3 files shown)


hjoliver (approved):

hjoliver:
Good question. I guess there's no other way of getting there, other than via `cylc install`?

wxtim:
> Good question. I guess there's no other way of getting there, other than via `cylc install`?

Is there anything to stop a user manually creating this file?

MetRonnie:
What if the user deliberately left the file there to get cylc-rose to pick up the options?

wxtim:
> What if the user deliberately left the file there to get cylc-rose to pick up the options?

Shouldn't be doing that - it's a reserved filename.  (I should double check that we throw an exception if they add it manually to cylc-src!?)

The legit case is where one is using `cylc re-install` and there's a swathe of logic for that.

MetRonnie:
Ah but hang on a minute, if the run dir already exists then `cylc install` itself will fail with
```
WorkflowFilesError: '~/cylc-run/tmpflow' already exists
To reinstall, use `cylc reinstall`
```

wxtim:
> Ah but hang on a minute, if the run dir already exists then `cylc install` itself will fail with
> 
> ```
> WorkflowFilesError: '~/cylc-run/tmpflow' already exists
> To reinstall, use `cylc reinstall`
> ```
> 
> Unless I am missing something?

No, I don't think you are - I'm trying to work out if we should keep this because it covers an eventuality like the one Hilary came across, or just ditch it because it should never happen.

wxtim:
Closed on the basis that the circumstance which cause the issue to arise amount to a bug in Cylc and therefore should not be handled.

PR #146, opened 2022-06-29
jinja2: fix parsing of zeros

oliver-sanders:
Follow on from #140 which introduced back support for integers with leading zeros (e.g. `01`, `002`, etc).

Unfortunately the zero-stripping was a little over zealous so would strip the zero from the number `0` :facepalm:.


datamel (approved):
Thanks @oliver-sanders, checked out and tested as working.

wxtim (approved):

PR #148, opened 2022-06-29
Fix.deprecation message about suiterc section

wxtim:

## Summary
1. Warnings referring only to Cylc 8 should not appear in Cylc 7 compatibility mode.
2. Warning about `[jinja2:suite.rc]` should make it clear that they refer to `rose-suite.conf` and use the section syntax consistently.

## Requirements check-list

- [x] No change log entry required (Change to user messages not functionality).

### DIFF
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -25,7 +25,7 @@ from typing import TYPE_CHECKING, Union
 
 from cylc.flow.hostuserutil import get_host
 from cylc.flow import LOG
-from cylc.flow.flags import cylc7_back_compat
+import cylc.flow.flags as flags
 from cylc.rose.jinja2_parser import Parser, patch_jinja2_leading_zeros
 from metomi.rose import __version__ as ROSE_VERSION
 from metomi.isodatetime.datetimeoper import DateTimeOperator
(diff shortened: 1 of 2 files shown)


cylc/rose/utilities.py
```diff
@@ -25,6 +25,7 @@
 
 from cylc.flow.hostuserutil import get_host
 from cylc.flow import LOG
+from cylc.flow.flags import cylc7_back_compat
```
MetRonnie, line 28: (In practice this works. However it does make a copy of the flag instead of using the flag directly, so I would be tempted to instead do `import cylc.flow.flags.cylc7_back_compat`)

wxtim: I'm not sure that it matters, except in terms of using a trivially small amount of extra memory?

datamel (approved):
No problems found from me. Thanks @wxtim.

cylc/rose/utilities.py
MetRonnie: In principle a flag should be used directly in case its value changes elsewhere, I think. But doesn't affect this particular flag as it shouldn't change once set to `True`

PR #166, opened 2022-08-09
Unset environment variable set in doctests.

wxtim:

## Requirements check-list

- [x] Does not need tests: is itself a test change
- [x] No change log entry required (test change).

Suggest @MetRonnie  be single reviewer because 
- change is tiny
- you found fix independently

### DIFF
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -465,6 +465,7 @@ def merge_opts(config, opt_conf_keys):
         Merge all three options.
         >>> merge_opts(conf, 'gimmel')
         'aleph bet gimmel'
+        >>> os.environ['ROSE_SUITE_OPT_CONF_KEYS'] = ''
     """
     all_opt_conf_keys = []
     if 'opts' in config:
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -458,14 +458,18 @@ def merge_opts(config, opt_conf_keys):
         'aleph gimmel'
 
         Merge options from defines and environment.
-        >>> import os; os.environ['ROSE_SUITE_OPT_CONF_KEYS'] = 'bet'
-        >>> merge_opts(conf, '')
+        >>> from pytest import MonkeyPatch
+        >>> with MonkeyPatch.context() as mp:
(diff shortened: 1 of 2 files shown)


cylc/rose/utilities.py
```diff
@@ -465,6 +465,7 @@ def merge_opts(config, opt_conf_keys):
         Merge all three options.
         >>> merge_opts(conf, 'gimmel')
         'aleph bet gimmel'
+        >>> os.environ['ROSE_SUITE_OPT_CONF_KEYS'] = ''
```
MetRonnie, line 468: I would move this bit with the env vars out into a unit test where you can monkeypatch, rather than mess with the environment at all.

Or, it turns out you can use the monkeypatch context manager
```python-console
        Merge options from defines and environment.
        >>> from pytest import MonkeyPatch
        >>> with MonkeyPatch.context() as mp:
        ...     mp.setenv('ROSE_SUITE_OPT_CONF_KEYS', 'bet')
        ...     merge_opts(conf, '')
        'aleph bet'

        Merge all three options.
        >>> with MonkeyPatch.context() as mp:
        ...     mp.setenv('ROSE_SUITE_OPT_CONF_KEYS', 'bet')
        ...     merge_opts(conf, 'gimmel')
        'aleph bet gimmel'
```

tests/functional/test_rose_stem.py
```diff
@@ -244,7 +244,6 @@ def _inner_fn(rose_stem_cmd, verbosity=verbosity):
             **setup_stem_repo
         }
     yield _inner_fn
-    shutil.rmtree(setup_stem_repo['suite_install_dir'])
```
MetRonnie, line 247: I'm not sure this should be removed. It's a workflow run dir, right (not in `temp_path`)? So doesn't get cleaned up any other way unless I'm missing something?

wxtim: stet 
```suggestion
     shutil.rmtree(setup_stem_repo['suite_install_dir']) 
```

cylc/rose/utilities.py
wxtim: I really want to have the example in the code, so I'll try the monkeypatching. :)

MetRonnie (approved):

PR #167, opened 2022-08-09
Improve test cleanup

wxtim:

### DIFF
--- a/tests/functional/test_rose_stem.py
+++ b/tests/functional/test_rose_stem.py
@@ -172,6 +172,7 @@ def setup_stem_repo(tmp_path_factory, monkeymodule):
         'suitename': suitename,
         'suite_install_dir': suite_install_dir
     }
+    shutil.rmtree(suite_install_dir)
 
 
 @pytest.fixture(scope='class')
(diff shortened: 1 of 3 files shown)


wxtim:
> I think this has duplicate changes from #166, do you want to rebase on master and whittle those down?

Should be fine now.

tests/functional/test_rose_stem.py
```diff
@@ -172,6 +172,7 @@ def setup_stem_repo(tmp_path_factory, monkeymodule):
         'suitename': suitename,
         'suite_install_dir': suite_install_dir
     }
+    shutil.rmtree(suite_install_dir)
```
MetRonnie, line 175: is it worth only removing if the tests passed, otherwise leaving it behind so you can have a poke around? This is how we check in cylc-flow integration tests:
https://github.com/cylc/cylc-flow/blob/82bc2c9c9d38fd339f75a84e02b087e907cbc07b/tests/integration/conftest.py#L73-L80

wxtim: ... and having learned how to do that... I shall create [another issue](#168 ) to do that elsewhere in testing.

MetRonnie (approved):

PR #170, opened 2022-08-22
Rose stem was not handling no arguments correctly

wxtim:
## Issue description

(not documented elsewhere)

The rose stem tests I translated from Rose 2019

Rose Stem 2019 did not accept _any_ arguments ([docs link](http://metomi.github.io/rose/2019.01.2/html/tutorial/rose/furthertopics/rose-stem.html#the-source-argument)). It assumed that users would provide a folder containing a rose-stem workflow either by running `rose stem` with now `--source` opts in a directory where `$PWD/rose-stem` was a valid rose-stem suite, or that the first `--source` option was such a suite. If users wanted to put the workflow elsewhere without adding it to the sources list they could use the rose `-C` argument. 

A user discovered that the conversion function was not getting the workflow config dir from the first source correctly.

## Requirements check-list

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,9 @@ ones in. -->
 
 ### Fixes
 
+[#170](https://github.com/cylc/cylc-rose/pull/170) - Fix support for
+installing workflows using --source.
+
 [#140](https://github.com/cylc/cylc-rose/pull/140) -
 Support integers with leading zeros (e.g `001`) to back support Rose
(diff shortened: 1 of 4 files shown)


datamel (commented):
Update from me, checked out and tested following cases:
- rose stem no args, working fine
- rose stem --source, working fine
- rose stem absolute path, not currently working for me, @wxtim aware aware and trying to reproduce.

PR #172, opened 2022-08-26
Rose Stem Fixes

wxtim:

## Summary of issue

At Rose 2019  the command `rose stem  --source=X=$PWD` would set the variable `conf_dir` to None by default, before passing the options to the `StemRunner.process` method.

At Rose 2 I had failed to do this and as a result attempts to calculate the `conf_dir` in Rose failed because there was no `conf_dir` to reference.

n.b. In the process of translating the Rose 2019 code I renamed `conf_dir` and `source` to `source` and `stem_sources` respectively. I'm not completely happy with that decision, but I don't want to add a reversion to it to this PR.

## Additional things

- I've copied the Rose 2019 help docs for rose stem - they probably aren't perfect, but I've created https://github.com/cylc/cylc-rose/issues/176 to improve them.
- A bug with the target for installation of non-fcm workflows has been written up as #177 

## Requirements check-list

### DIFF
--- a/cylc/rose/stem.py
+++ b/cylc/rose/stem.py
@@ -23,25 +23,53 @@ Install a rose-stem suite using cylc install.
 
 To run a rose-stem suite use "cylc play".
 
-Looks for a suite to install either in $PWD/rose-stem, or in a specified
-location - e.g:
+Install a suitable suite with a specified set of source tree(s).
 
-# Install a rose-stem suite from PWD/rose-stem
-$ rose stem
+Default values of some of these settings are suite-dependent, specified
+in the `rose-suite.conf` file.
 
-# Install a rose-stem suite from PWD/relative-path.
-$ rose stem relative-path
+Examples
 
-# Install a rose-stem suite from specified abs path.
-$ rose stem /absolute/path
+    rose stem --group=developer
+    rose stem --source=/path/to/source --source=/other/source --group=mygroup
+    rose stem --source=foo=/path/to/source --source=bar=fcm:bar_tr@head
+
+Jinja2 Variables
+
+    Note that `<project>` refers to the FCM keyword name of the repository in
+    upper case.
+
+    HOST_SOURCE_<project>
+        The complete list of source trees for a given project. Working copies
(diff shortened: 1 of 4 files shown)


datamel (approved):
Checked out locally, tests run and compared this behaviour on master. This works as detailed in the description. Thanks @wxtim.

datamel:
Just to note that this is still the case: https://github.com/cylc/cylc-rose/pull/170#pullrequestreview-1083485279

So, either: 
- As this was not supported in previous version, the help needs updating as it currently reads:
``` 
# Install a rose-stem suite from PWD/relative-path.
$ rose stem relative-path  #<<<< This does not work

# Install a rose-stem suite from specified abs path.
$ rose stem /absolute/path  #<<<< This does not work
```
- or, absolute path needs supporting/fixing.

wxtim:
@wxtim  - Fix the help docs

datamel (approved):
I'm happy with the updated help. The comment about the absolute path causing error is only the case for non-svn sources and so the help update resolves that comment from me. Thanks @wxtim.

cylc/rose/stem.py
```diff
@@ -23,25 +23,53 @@
...
-# Install a rose-stem suite from specified abs path.
-$ rose stem /absolute/path
+Install a suitable suite with a specified set of source tree(s).
+
+Default values of some of these settings are suite-dependent, specified
+in the `rose-suite.conf` file.
+
+Examples
+
+    rose stem --group=developer
```
datamel, line 33: Do you think out users will understand the --group option? It used to appear in the help and is now not explained?

wxtim: You've spotted an aspect of copying the Rose 2019 help that I forgot. 😞

wxtim: I've re-added the help for those things - I've also put the rose-stem specific options into a group so that they get their own sub heading.

dpmatthews (approved):
Fixes the problem in my tests

datamel (approved):
Thanks @wxtim

Issue #157, opened 2022-07-26
coverage: get coverage up


oliver-sanders:
We should be able to hit 100% coverage for this project.

**Pull requests welcome!**

MetRonnie:
`stem.py` is dragging down coverage. https://app.codecov.io/gh/cylc/cylc-rose

oliver-sanders:
I think the reason for that is because the Rose Stem tests invoke `rose stem` via the CLI so coverage is not capturing the test usage.

https://github.com/cylc/cylc-rose/blob/9b1ffd08c46d5d254cb581c766bb98528e2ba82c/tests/functional/test_rose_stem.py#L205-L210

Luckily `rose stem` wraps Cylc, and Cylc has a system for auto-capturing coverage for CLI calls.

```
$ export CYLC_COVERAGE=2
$ pytest
$ cp ../cylc-flow/<coverage-file-pattern> .
$ coverage combine
```

wxtim:
The scope of the existing functional tests (Which I transferred to using Pytest but otherwise left well alone) is to check the contents of the `rose-suite-cylc-install.conf` for the variables set by rose-stem. This approach offers no tests for 

- Workflow installation name.
- Any cases where source is not an svn repo.
- The interaction of Rose and Rose-stem CLI options (e.g. `rose stem --group="Foo" -S='RUN_NAMES="Foo"'`)

wxtim:
I've closed my attempt to answer this (https://github.com/cylc/cylc-rose/pull/196), but that may contain stuff worth pulling out.

PR #181, opened 2022-09-13
Improve rose stem coverage

wxtim:
These changes partially address #157 

I have added some unit tests designed to cover holes in the functional testing for `rose stem` I translated from Rose 2019.

added @dpmatthews  as a functional review WRT mocking subprocessed fcm commands.

## Requirements check-list

- [x] Test improvements - No change log, documentation change &c

### DIFF
--- a/cylc/rose/stem.py
+++ b/cylc/rose/stem.py
@@ -153,7 +153,9 @@ class RoseStemVersionException(Exception):
     def __init__(self, version):
         Exception.__init__(self, version)
         if version is None:
-            self.suite_version = "not rose-stem compatible"
+            self.suite_version = (
+                "does not have ROSE_VERSION set in the rose-suite.conf"
+            )
         else:
             self.suite_version = "at version %s" % (version)
 
@@ -225,24 +227,32 @@ class StemRunner:
 
     def __init__(self, opts, reporter=None, popen=None, fs_util=None):
         self.opts = opts
+
         if reporter is None:
             self.reporter = Reporter(opts.verbosity - opts.quietness)
         else:
             self.reporter = reporter
+
         if popen is None:
             self.popen = RosePopener(event_handler=self.reporter)
         else:
             self.popen = popen
+
         if fs_util is None:
             self.fs_util = FileSystemUtil(event_handler=self.reporter)
         else:
             self.fs_util = fs_util
+
         self.host_selector = HostSelector(event_handler=self.reporter,
                                           popen=self.popen)
 
     def _add_define_option(self, var, val):
-        """Add a define option passed to the SuiteRunner."""
+        """Add a define option passed to the SuiteRunner.
 
+        Args:
+            var: Name of variable to set
+            val: Value of variable to set
+        """
         if self.opts.defines:
             self.opts.defines.append(SUITE_RC_PREFIX + var + '=' + val)
         else:
@@ -250,7 +260,7 @@ class StemRunner:
         self.reporter(ConfigVariableSetEvent(var, val))
         return
 
-    def _get_base_dir(self, item):
+    def _get_fcm_loc_layout_info(self, src_tree):
         """Given a source tree return the following from 'fcm loc-layout':
            * url
            * sub_tree
@@ -259,15 +269,17 @@ class StemRunner:
            * project
         """
 
-        ret_code, output, stderr = self.popen.run('fcm', 'loc-layout', item)
+        ret_code, output, stderr = self.popen.run(
+            'fcm', 'loc-layout', src_tree)
         if ret_code != 0:
-            raise ProjectNotFoundException(item, stderr)
+            raise ProjectNotFoundException(src_tree, stderr)
 
         ret = {}
         for line in output.splitlines():
             if ":" not in line:
                 continue
             key, value = line.split(":", 1)
+
             if key and value:
                 ret[key] = value.strip()
 
@@ -289,7 +301,8 @@ class StemRunner:
                     break
         return project
 
-    def _deduce_mirror(self, source_dict, project):
+    @staticmethod
+    def _deduce_mirror(source_dict, project):
         """Deduce the mirror location of this source tree."""
 
         # Root location for project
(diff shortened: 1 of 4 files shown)


cylc/rose/stem.py
```diff
@@ -259,15 +269,17 @@ def _get_base_dir(self, item):
...
-            raise ProjectNotFoundException(item, stderr)
+            raise ProjectNotFoundException(src_tree, stderr)
 
         ret = {}
         for line in output.splitlines():
             if ":" not in line:
                 continue
             key, value = line.split(":", 1)
+
             if key and value:
```
wxtim, line 283: @dpmatthews  - This is splitting lines from `fcm loc-layout`. Will it ever be false?

wxtim:
Coverage identifies the following remaining holes:
- 112,124,178 - Some string output from "events" and Exceptions is untested.
- 283 - A possible guard against a subprocessed FCM command going wrong which never gets hit.
- 362 - I'd need to dig a lot further into FCM to meaningfully test this. 
- 491 - Splitting standard items from config doesn't happen if there is not exactly 1 `=` in them - Is this sensible? 
- 606 - This looks trivial - I'm not sure it needs testing, but it's easy to test if you want it to be. What do people think?

tests/unit/test_rose_stem_units.py
```diff
@@ -54,3 +73,225 @@ def test_get_source_opt_from_args(tmp_path, monkeypatch, args, expect):
...
+@pytest.mark.parametrize(
+    'attribute, object_, kwargs', [
+        param('reporter', Reporter, {}, id='reporter default'),
+        param('reporter', str, {'reporter': 'foo'}, id='reporter set'),
+        param('popen', RosePopener, {}, id='popen default'),
+        param('popen', str, {'popen': 'foo'}, id='popen set'),
+        param('fs_util', FileSystemUtil, {}, id='fs_util default'),
+        param('fs_util', str, {'fs_util': 'foo'}, id='fs_util set')
+    ]
+)
```
oliver-sanders, line 101: This creates one StemRunner per test run which isn't necessary.

Ok to ignore but IMO this would be better/clearer without parametrize. If one of the subtests goes wrong it is easier to work out what's broken when the traceback shows something like:

```
assert isinstance(stem_runner.reporter, Reporter)
Traceback ...
AssertionError: NoneType != Reporter
```

Rather than `assert isinstance(item, object_)`, then having to find the parametrised part of the test name that failed, then get into the source code to lookup the values of `item` and `object_` to find out what actually went wrong.

setup.cfg
```diff
@@ -55,8 +55,8 @@ packages = find_namespace:
 python_requires = >=3.7
 include_package_data = True
 install_requires =
-    metomi-rose==2.1.*
-    cylc-flow==8.1.*
+    # metomi-rose==2.*.*
+    # cylc-flow==8.*.*
```
MetRonnie, line 59: ❓

MetRonnie (changes requested):

setup.cfg
wxtim: Accidentally committed.
```suggestion
    metomi-rose==2.1.*
    cylc-flow==8.1.*
```

tests/unit/test_rose_stem_units.py
```diff
@@ -54,3 +73,225 @@ def test_get_source_opt_from_args(tmp_path, monkeypatch, args, expect):
...
+
+    if expect:
+        (stem_suite_subdir / 'rose-suite.conf').write_text(
+            'ROSE_STEM_VERSION=1')
+        stemrunner = get_StemRunner({}, {'stem_sources': stem_sources})
+        assert stemrunner._this_suite() == str(stem_suite_subdir)
+    else:
+        with pytest.raises(RoseSuiteConfNotFoundException):
+            stemrunner = get_StemRunner({}, {'stem_sources': stem_sources})
+            assert stemrunner._this_suite() == str(stem_suite_subdir)
```
oliver-sanders, line 263: This assert doesn't get run right?

MetRonnie (approved):
Most of this went over my head but good to see coverage up by 16% points

tests/unit/test_rose_stem_units.py
wxtim: Mea Cupa

tests/unit/test_rose_stem_units.py
```diff
@@ -54,3 +73,225 @@ def test_get_source_opt_from_args(tmp_path, monkeypatch, args, expect):
...
+def test__this_suite(
+    get_StemRunner, monkeypatch, tmp_path, stem_sources, expect
+):
+    """It returns a sensible suite-dir."""
+    stem_suite_subdir = tmp_path / 'rose-stem'
+    stem_suite_subdir.mkdir()
+
+    if stem_sources == 'infer':
+        stem_sources = []
+        import cylc
```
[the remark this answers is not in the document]
wxtim: Used by the mock, needs moved to the top.

tests/unit/test_rose_stem_units.py
wxtim: Correct.

wxtim:
> Most of this went over my head but good to see coverage up to 94%

Some of the methods which act as interfaces to FCM are pretty opaque to me, and there are at least 2 lines where I had to put in considerable effort to ensure that there wasn't a bug.

tests/unit/test_rose_stem_units.py
```diff
@@ -54,3 +74,220 @@ def test_get_source_opt_from_args(tmp_path, monkeypatch, args, expect):
...
+
+    if expect:
+        (stem_suite_subdir / 'rose-suite.conf').write_text(
+            'ROSE_STEM_VERSION=1')
+        stemrunner = get_StemRunner({}, {'stem_sources': stem_sources})
+        assert stemrunner._this_suite() == str(stem_suite_subdir)
+    else:
+        with pytest.raises(RoseSuiteConfNotFoundException):
+            stemrunner = get_StemRunner({}, {'stem_sources': stem_sources})
+            stemrunner._this_suite()
```
MetRonnie, line 259: If the exception is expected on the second line, then suggest moving the first out of there
```suggestion
        stemrunner = get_StemRunner({}, {'stem_sources': stem_sources})
        with pytest.raises(RoseSuiteConfNotFoundException):
            stemrunner._this_suite()
```

oliver-sanders (approved):

PR #189, opened 2022-10-04
20221004 t0858  1.1.x=>test lower case jinja2 bools

wxtim:

A user was using `true` and `false` as literals in their `[template variables]` section of `rose-suite.conf`. It appears that Jinja2 handles this. I added a test so that we can track this functionality.

## Requirements check-list

This PR adds scope to a test, reflecting investigation into customer usage:
  - [x] Does not need tests.
  - [x] No change log entry required.


oliver-sanders:
Was thinking that this test should be added to the Jinja2 variable parser.

Took a look and it appears that it already is :facepalm: 

https://github.com/cylc/cylc-rose/blob/f4398efbe641fc80e3907a5190e47962b886b22f/cylc/rose/jinja2_parser.py#L239-L240

PR #196, opened 2022-11-25
100% cov

### LINKED ISSUE #157: coverage: get coverage up

wxtim:
Opened PR to allow commenting on code changes.

Requires https://github.com/cylc/cylc-flow/pull/5246 to pass all tests

These changes close #157 

## Requirements check-list

- [x] Already covered by existing tests.
- [x] No change log entry required (invisible to users).

### DIFF
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -71,8 +71,7 @@ def post_install(srcdir=None, opts=None, rundir=None):
         srcdir=srcdir, opts=opts, rundir=rundir
     )
     # Finally dump a log of the rose-conf in its final state.
-    if results['fileinstall']:
-        dump_rose_log(rundir=rundir, node=results['fileinstall'])
+    dump_rose_log(rundir=rundir, node=results['fileinstall'])
 
     return results
 
@@ -177,11 +176,6 @@ def record_cylc_install_options(
     """
     # Create a config based on command line options:
     cli_config = get_cli_opts_node(opts, srcdir)
-    # Leave now if there is nothing to do:
-    if not cli_config:
-        return False
-    # raise error if CLI config has multiple templating sections
-    identify_templating_section(cli_config)
 
     # Construct path objects representing our target files.
     (Path(rundir) / 'opt').mkdir(exist_ok=True)
@@ -218,8 +212,7 @@ def record_cylc_install_options(
     dumper.dump(cli_config, str(conf_filepath))
 
     # Merge the opts section of the rose-suite.conf with those set by CLI:
-    if not rose_conf_filepath.is_file():
-        rose_conf_filepath.touch()
+    rose_conf_filepath.touch()
     rose_suite_conf = loader.load(str(rose_conf_filepath))
     rose_suite_conf = add_cylc_install_to_rose_conf_node_opts(
         rose_suite_conf, cli_config
@@ -245,7 +238,6 @@ def rose_fileinstall(srcdir=None, opts=None, rundir=None):
 
     # Load the config tree
     config_tree = rose_config_tree_loader(rundir, opts)
-
     if any(i.startswith('file') for i in config_tree.node.value):
         try:
             startpoint = os.getcwd()
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -257,7 +257,6 @@ def rose_fileinstall(srcdir=None, opts=None, rundir=None):
             config_tree.node = config_tree.node.set(
                 keys=["file-install-root"], value=str(rundir)
             )
-
             # Artificially set rose to verbose.
             event_handler = Reporter(3)
             fs_util = FileSystemUtil(event_handler)
--- a/cylc/rose/stem.py
+++ b/cylc/rose/stem.py
@@ -253,8 +253,11 @@ class StemRunner:
            * peg_rev
            * root
            * project
-        """
 
+        Simple parser of output expected to be in the format:
+
+        Key: Value.
+        """
         ret_code, output, stderr = self.popen.run(
             'fcm', 'loc-layout', src_tree)
         if ret_code != 0:
@@ -265,7 +268,6 @@ class StemRunner:
             if ":" not in line:
                 continue
             key, value = line.split(":", 1)
-
             if key and value:
                 ret[key] = value.strip()
 
@@ -421,6 +423,16 @@ class StemRunner:
             url = self.host_selector.get_local_host() + ':' + url
         return url
 
+    def _parse_auto_opts(self):
+        auto_opts = self._read_auto_opts()
+        if auto_opts:
+            automatic_options = auto_opts.split()
+            for option in automatic_options:
+                elements = option.split("=")
+                if len(elements) == 2:
+                    self._add_define_option(
+                        elements[0], '"' + elements[1] + '"')
+
     def process(self):
         """Process STEM options into 'rose suite-run' options."""
         # Generate options for source trees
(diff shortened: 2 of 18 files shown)


cylc/rose/utilities.py
```diff
@@ -128,54 +128,43 @@ def get_rose_vars_from_config_node(config, config_node, environ):
...
-            item[0][1]: item[1].value for item in
-            config_node.value['template variables'].walk()
-            if item[1].state == ConfigNode.STATE_NORMAL
-        }
+    config['env'] = {
+        item[0][1]: item[1].value for item in
+        config_node.value['env'].walk()
+        if item[1].state == ConfigNode.STATE_NORMAL
+    }
+    config['template_variables'] = {
```
wxtim, line 136: templating will always be in config_node.value, it's set around line 77, as is `config['env']`

cylc/rose/utilities.py
```diff
@@ -128,54 +128,43 @@ def get_rose_vars_from_config_node(config, config_node, environ):
...
-                        )
-                    except Exception:
-                        raise ConfigProcessError(
-                            [templating, key],
-                            value,
-                            f'Invalid template variable: {value}'
-                            '\nMust be a valid Python or Jinja2 literal'
-                            ' (note strings "must be quoted").'
-                        ) from None
+    with patch_jinja2_leading_zeros():
```
wxtim, line 144: We'd never use this condition

cylc/rose/utilities.py
```diff
@@ -391,9 +380,9 @@ def get_cli_opts_node(opts=None, srcdir=None):
         )
 
     # Specialised treatement of optional configs.
-    if 'opts' not in newconfig:
-        newconfig['opts'] = ConfigNode()
```
wxtim, line 395: Never False

cylc/rose/entry_points.py
```diff
@@ -71,8 +71,7 @@ def post_install(srcdir=None, opts=None, rundir=None):
         srcdir=srcdir, opts=opts, rundir=rundir
     )
     # Finally dump a log of the rose-conf in its final state.
-    if results['fileinstall']:
```
wxtim, line 74: We set `results['fileinstall']` at line 70 - this condition will never be false.

cylc/rose/entry_points.py
```diff
@@ -177,9 +176,6 @@ def record_cylc_install_options(
     """
     # Create a config based on command line options:
     cli_config = get_cli_opts_node(opts, srcdir)
-    # Leave now if there is nothing to do:
-    if not cli_config:
```
wxtim, line 181: `get_cli_opts_node` will always return a `cli_config`.

cylc/rose/entry_points.py
```diff
@@ -215,8 +211,7 @@ def record_cylc_install_options(
     dumper.dump(cli_config, str(conf_filepath))
 
     # Merge the opts section of the rose-suite.conf with those set by CLI:
-    if not rose_conf_filepath.is_file():
```
wxtim, line 218: There is a check for this right at the beginning of the plugin's work. We shouldn't need to check again here.

cylc/rose/stem.py
```diff
@@ -102,18 +102,6 @@ def __repr__(self):
     __str__ = __repr__
 
 
-class ConfigSourceTreeSetEvent(Event):
```
wxtim, line 105: I don't know what used to use this, but it's gone!

cylc/rose/entry_points.py
oliver-sanders: Agreed.

cylc/rose/entry_points.py
oliver-sanders: `touch` doesn't fail if the file is present so skipping the check is ok.

cylc/rose/utilities.py
```diff
@@ -33,7 +33,7 @@
 from metomi.rose.config_processor import ConfigProcessError
 from metomi.rose.env import env_var_process, UnboundEnvironmentVariableError
 
-if TYPE_CHECKING:
+if TYPE_CHECKING:  # pragma: no cover
```
oliver-sanders, line 36: There's a better solution to this, we can configure patterns for coverage to exclude.

Here's how we skip these lines automatically in Cylc:

https://github.com/cylc/cylc-flow/blob/master/.coveragerc#L24-L39

```
exclude_lines =
    pragma: no cover


    # Don't complain if tests don't hit defensive assertion code:
    raise NotImplementedError
    return NotImplemented


    # Ignore type checking code:
    if (typing\.)?TYPE_CHECKING:
    @overload( |$)


    # Don't complain about ellipsis (exception classes, typing overloads etc):
    \.\.\.


    # Ignore abstract methods
    @(abc\.)?abstractmethod
```

cylc/rose/utilities.py
oliver-sanders: Agreed.

oliver-sanders: Lines 66-67 can be removed too.

cylc/rose/utilities.py
oliver-sanders: Agreed.

cylc/rose/utilities.py
oliver-sanders: I think you're right but it might be worth adding an explicit `set` after the config node is created because I think this is relying on assumptions about automatic variables being set in all situations. An assumption which could potentially break down the line.

tests/functional/test_pre_configure.py
```diff
@@ -37,26 +36,22 @@ def envar_exporter(dict_):
...
 )
-def test_validate_fail(srcdir, expect, cylc_validate_cli):
+def test_validate_fail(srcdir, cylc_validate_cli):
     srcdir = Path(__file__).parent / srcdir
-    validate = cylc_validate_cli(srcdir)
-    assert validate.ret == 1
-    if expect:
-        assert re.findall(expect, str(validate.exc))
+    result = cylc_validate_cli(str(srcdir))
+    assert result.ret != 0
```
oliver-sanders, line 54: This seems a little complicated. Rather than running these scripts through something which emulates `Popen` then checking the artificial return code to determine whether it failed, why not run the check directly e.g:

```python
def test_validate_fail(srcdir):
    srcdir = Path(__file__).parent / srcdir
    with pytest.raises(Exception):
        validate(src_dir)
```

That way the tests can interact with the code directly, catch/expect the exceptions by name, mock/path, caplog, capsys as appropriate.

tests/conftest.py
```diff
@@ -181,43 +133,71 @@ def _inner(workflow_id, opts=None):
...
+    )
 
 
 @pytest.fixture(scope='module')
 def mod_cylc_install_cli(mod_capsys, mod_caplog):
-    return _cylc_install_cli(mod_capsys, mod_caplog)
+    return _cylc_cli(
+        mod_capsys, mod_caplog,
+        cylc_install, install_gop
+    )
```
oliver-sanders, line 166: Maybe do this more similarly to the Cylc integration tests.

tests/functional/test_pre_configure.py
```diff
@@ -118,6 +113,8 @@ def test_process(tmp_path, srcdir, envvars, args):
         capture_output=True,
         env=envvars
     ).stdout.decode()
+    # TODO: Fix this - it fails in a way I don't understand.
+    # result = cylc_view_cli(str(srcdir), {'process': True})
```
oliver-sanders, line 117: TODO

oliver-sanders (commented):
:100:!

Great, glad to have full coverage for at least one of our repos.

tests/conftest.py
wxtim: This might be too long ago, and I'm not terribly familiar with the Cylc Integration tests... but...

My impression is that they largely revolve around running flows which until recently were written to the flow directory.

Here most of the tests revolve around the install process, so I'm not sure that the same approach makes sense.

wxtim:
@oliver-sanders  - this one is by no means a priority, but I think it's worth getting in at some point.

tests/conftest.py
oliver-sanders: I think my point was that this implementation differs from the one you are currently adding in cylc-flow with the `cylc vr` pull request. It would make sense to do it the same way for both.

wxtim:
Thoroughly stale: I've made a note on my personal todo list that this branch is worth pulling stuff out of.

PR #208, opened 2023-02-02
fix flake 8 breakage

wxtim:
Fix automated tests:
- Fix a flake8 issue
- Fix an SQL issue caused by more stringent database checks.

## Requirements check-list

- [x] Already covered by existing tests.
- [x] No change log entry required - changes to internal checks and tests

### DIFF
--- a/tests/unit/test_config_node.py
+++ b/tests/unit/test_config_node.py
@@ -153,9 +153,9 @@ def test_add_cylc_install_to_rose_conf_node_opts(rose_conf, cli_conf, expect):
     expect_opt += ' (cylc-install)'
 
     assert result.comments == [(
-        f' Config Options \'{expect_opt}\' from CLI '
-        'appended to options '
-        'already in `rose-suite.conf`.'
+        f'Config Options \'{expect_opt}\' from CLI'
+        ' appended to options'
+        ' already in `rose-suite.conf`.'
     )]
     assert result.state == ''
 
--- a/tests/unit/test_config_node.py
+++ b/tests/unit/test_config_node.py
@@ -153,7 +153,7 @@ def test_add_cylc_install_to_rose_conf_node_opts(rose_conf, cli_conf, expect):
     expect_opt += ' (cylc-install)'
 
     assert result.comments == [(
-        f'Config Options \'{expect_opt}\' from CLI'
+        f' Config Options \'{expect_opt}\' from CLI'
         ' appended to options'
         ' already in `rose-suite.conf`.'
     )]
--- a/tests/unit/test_platform_utils.py
+++ b/tests/unit/test_platform_utils.py
@@ -22,7 +22,6 @@ import os
 import pytest
 from pathlib import Path
 from shutil import rmtree
-from subprocess import run
 import sqlite3
 from uuid import uuid4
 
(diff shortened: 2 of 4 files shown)


tests/unit/test_platform_utils.py
```diff
@@ -146,22 +146,32 @@ def fake_flow():
...
+        "INSERT INTO task_jobs (cycle, name, submit_num, platform_name)"
+        "    VALUES ('2', 'baz', 1, 'milk');",
+        "CREATE TABLE workflow_params(key TEXT, value TEXT);",
+        "INSERT INTO workflow_params"
+        f"    VALUES ('cylc_version', {cylc_version!r});",
+        "CREATE TABLE workflow_template_vars(key TEXT, value TEXT);",
+    ):
+        cursor.execute(line)
+    conn.commit()
+    conn.close()
```
MetRonnie, line 174: Instead of using an sqlite3 connection directly, could import and use
```py
with CylcWorkflowDAO(db_file, create_tables=True) as dao:
    dao.connect().execute(
```
and with that can get rid of the `CREATE TABLES` statements

wxtim: Yes, but I don't believe that it should block this.

datamel (approved):

cylc/rose/stem.py
```diff
@@ -339,7 +339,7 @@ def _ascertain_project(self, item):
             item = os.path.abspath(os.path.join(os.getcwd(), item))
 
         if project is not None:
-            print(f"[WARN] Forcing project for '{item}' to be '{project}'")
+            print(f"[WARN] Forcing project for {item!r} to be {project!r}")
```
oliver-sanders, line 342: IMO it's clearer what `'{foo}'` means as compared to `{foo!r}` on the basis that reusing simple constructs is simpler/lower-barrier-to-entry than learning exotic syntaxes.

I don't mind much either way but just to raise the point (as this will also hit our other repos) that I don't think it's worth the time/effort of converting a larger codebase (e.g. cylc-flow) to this pattern.

datamel: I'm also happy for this to be a global ignore then for code readability.

wxtim: I agree - I've added a tox.ini

tox.ini
```diff
@@ -0,0 +1,23 @@
...
+    B028
+
+exclude=
+    build,
+    dist,
+    .git,
+    __pycache__,
+    .tox,
+paths =
+    ./cylc/flow
```
oliver-sanders, line 22: I think this turns linting off:

```suggestion
    ./cylc/rose
```

(please test that linting still works after this change with no extra ages though!)

tests/unit/test_platform_utils.py
MetRonnie: On the contrary, I think using the DAO-created tables means the test better reflects the ground truth of what the DB is actually like, and more likely to catch problems with DB changes

tests/unit/test_config_node.py
```diff
@@ -153,7 +153,7 @@ def test_add_cylc_install_to_rose_conf_node_opts(rose_conf, cli_conf, expect):
     expect_opt += ' (cylc-install)'
 
     assert result.comments == [(
-        f' Config Options \'{expect_opt}\' from CLI '
+        f' Config Options {expect_opt!r} from CLI '
```
MetRonnie, line 156: In `tox.ini` you said this doesn't work at Python 3.7, but the tests are passing?

wxtim: Comment misleading - I've reworded it:
Flake 8 bugbear only checks for this issue at Python > 3.7

MetRonnie: Happy to disable it if we can't be bothered with it but the fact it isn't checked at Python 3.7 doesn't seem like a good reason

MetRonnie (approved):

Issue #217, opened 2023-03-21
`PluginError` using `cylc install -D root-dir`


ColemanTom:
### Description

Tested in cylc-flow=8.0.3 only.

If you try to do a `cylc install` and have `root-dir=...` syntax in the command, carrying over from `pre v2 rose`, the install command will spit out a WARNING message, install the workflow, but then fail afterwards.

```
$ cylc install -D "root-dir{work}=somehost=/tmp" $PWD
2023-03-21T14:15:45+11:00 WARNING - You have set "rose-suite.conf[root-dir]", which is not supported at Cylc 8. Use `[install] symlink dirs` in global.cylc instead.
INSTALLED basic-workflow/run3 from /home/ColemanTom/roses/basic-workflow
PluginError: Error in plugin cylc.post_install.rose_opts
AttributeError: 'NoneType' object has no attribute 'groupdict'
```

### Reproducible Example

Run `cylc install -D "root-dir{work}=somehost=/tmp"` for any workflow. You will get a WARNING and then an exception is thrown by `cylc-rose`.

### Expected Behaviour

WARNING should be an ERROR and `cylc install` should fail before the workflow is installed.

wxtim:
@ColemanTom - Did you mean `cylc install -S "root-dir{work}=somehost=/tmp" $PWD` - I'm not sure that I'd expect `-D` to work, at least not in any useful way - I'd expect what you've done to be the equivelent of setting the `rose-suite.conf` as follows:

```toml
# rose-suite.conf
!opts = x, y, z
root-dir{work}=somehost=/tmp

[env]
...
```

Is that what you intend?

PR #221, opened 2023-04-13
Root dir fails

wxtim:
Closes https://github.com/cylc/cylc-rose/issues/217

Allows deprecated config items to cause a hard-fail as well as a warning on validation.

@oliver-sanders Do we want this mechanism to make `<template engine:flow.cylc>` a cause for error rather than warning, since it's not required for back compat  and I don't think that we want people using it?

**Check List**


oliver-sanders:
As per #217 I think we want this *not* to error if `root-dir` is defined, just to warn.

wxtim:
> As per #217 I think we want this _not_ to error if `root-dir` is defined. Rose should just warn that this obsolete configuration is present, then move on. Note that `root-dir` may be necessary for Cylc 7 interoperability.

Just re-read the issue - I've got the wrong end of the stick, but it's an easy fix.

cylc/rose/utilities.py
```diff
@@ -325,6 +325,72 @@ def merge_rose_cylc_suite_install_conf(old, new):
...
+        # Doesn't have a section:
+        match = re.match(
+            r'^(?P<state>!{0,2})(?P<key>.*)\s*=\s*(?P<value>.*)', define)
+        if match:
+            groupdict = match.groupdict()
+            keys = [groupdict['key'].strip()]
+        else:
+            # This seems like it ought to be an error,
+            # But behaviour is consistent with Rose 2019
+            # See:
```
MetRonnie, line 383: Missing a link after "See:"?

cylc/rose/utilities.py
```diff
@@ -325,6 +325,72 @@ def merge_rose_cylc_suite_install_conf(old, new):
...
+            return False
+
+    if match['state'] in ['!', '!!']:
+        LOG.warning(
+            'CLI opts set to ignored or trigger-ignored will be ignored.')
+        return False
+    else:
+        return (keys, match['value'], match['state'])
```
oliver-sanders, line 391: I don't think we should be failing for ignored configs, I'm not sure that warning is helpful so we could just:

```suggestion
    return (keys, match['value'], match['state'])
```

oliver-sanders (commented):
Should put in a test for this one.

wxtim:
> Should put in a test for this one.

Is the doctest not sufficient?

wxtim:
> > Should put in a test for this one.
> 
> Is the doctest not sufficient?

Poke @oliver-sanders 

I think that this is a bug that may delay some people upgrading.

MetRonnie:
I can't remember how to test this, can you provide a minimal example?

wxtim:
> I can't remember how to test this, can you provide a minimal example? Also is this about sections or settings?

`cylc install -D "root-dir{work}=somehost=/tmp"`

MetRonnie (approved):

oliver-sanders (approved):

PR #225, opened 2023-05-04
Raise error if `"="` not in define

### LINKED ISSUE #226: Error if `"=" not in define`
Replaces https://github.com/metomi/rose/issues/2691

If a user were to make a typo and do

```bash
CYLC_VERSION=7 rose suite-run -D 'FOO42'
# or
CYLC_VERSION=8 cylc vip -D 'FOO42'
```

No warning or error is raised despite this define being largely meaningless.

Should we warn when defines are not in the form `[section]key=value` or `key=value`?

wxtim:
Closes #226 

Built atop https://github.com/cylc/cylc-rose/pull/221

**Check List**

- [x] ~No documentation update req'd~

### DIFF
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -39,7 +39,6 @@ from cylc.rose.utilities import (
     get_cli_opts_node,
     add_cylc_install_to_rose_conf_node_opts,
 )
-from cylc.flow.flags import cylc7_back_compat
 from cylc.flow.hostuserutil import get_host
 
 
@@ -123,7 +122,7 @@ def get_rose_vars(srcdir=None, opts=None):
         return config
 
     # Check for definitely invalid defines
-    if opts and hasattr(opts, 'defines') and not cylc7_back_compat:
+    if opts and hasattr(opts, 'defines'):
         invalid_defines_check(opts.defines)
 
     # Load the raw config tree
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -31,6 +31,7 @@ from cylc.rose.utilities import (
     dump_rose_log,
     get_rose_vars_from_config_node,
     identify_templating_section,
+    invalid_defines_check,
     rose_config_exists,
     rose_config_tree_loader,
     merge_rose_cylc_suite_install_conf,
@@ -38,6 +39,7 @@ from cylc.rose.utilities import (
     get_cli_opts_node,
     add_cylc_install_to_rose_conf_node_opts,
 )
+from cylc.flow.flags import cylc7_back_compat
 from cylc.flow.hostuserutil import get_host
 
 
@@ -119,6 +121,10 @@ def get_rose_vars(srcdir=None, opts=None):
         ):
             raise NotARoseSuiteException()
         return config
+
+    # Check for definitely invalid defines
+    if opts and hasattr(opts, 'defines') and not cylc7_back_compat:
+        invalid_defines_check(opts.defines)
 
     # Load the raw config tree
     config_tree = rose_config_tree_loader(srcdir, opts)
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -25,11 +25,12 @@ from typing import TYPE_CHECKING, Any, List, Tuple, Union
 
 from cylc.flow.hostuserutil import get_host
 from cylc.flow import LOG
+from cylc.flow.exceptions import CylcError
 from cylc.flow.flags import cylc7_back_compat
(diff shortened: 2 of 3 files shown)


hjoliver:
> Closes #2691

Wrong Issue number?

wxtim:
> > Closes #2691
> 
> Wrong Issue number?

I originally wrote this up as a rose issue (because it seems like a rose problem).

cylc/rose/entry_points.py
```diff
@@ -120,6 +122,10 @@ def get_rose_vars(srcdir=None, opts=None):
             raise NotARoseSuiteException()
         return config
 
+    # Check for definately invalid defines
```
hjoliver, line 125:
```suggestion
    # Check for definitely invalid defines
```

cylc/rose/utilities.py
```diff
@@ -325,6 +330,103 @@ def merge_rose_cylc_suite_install_conf(old, new):
...
+
+        # No invalid defines
+        >>> invalid_defines_check(['foo=12'])
+    """
+    invalid_defines = []
+    for define in defines:
+        if parse_cli_defines(define) == 'Invalid':
+            invalid_defines.append(define)
+    if invalid_defines:
+        msg = 'Invalid Suite Defines (should contain an =)'
```
hjoliver, line 358: Are we still using "Suite" as far as Rose is concerned?

cylc/rose/utilities.py
```diff
@@ -325,6 +330,103 @@ def merge_rose_cylc_suite_install_conf(old, new):
...
+        # Doesn't have a section:
+        match = re.match(
+            r'^(?P<state>!{0,2})(?P<key>.*)\s*=\s*(?P<value>.*)', define)
+        if match:
+            groupdict = match.groupdict()
+            keys = [groupdict['key'].strip()]
+        else:
+            # This seems like it ought to be an error,
+            # But behaviour is consistent with Rose 2019
+            # See:
```
hjoliver, line 419: See what?

cylc/rose/utilities.py
```diff
@@ -325,6 +330,103 @@ def merge_rose_cylc_suite_install_conf(old, new):
...
+            return "Invalid"
+
+    if match['state'] in ['!', '!!']:
+        LOG.warning(
+            'CLI opts set to ignored or trigger-ignored will be ignored.')
```
hjoliver, line 424: Is that a thing? (Use the ignore syntax on the command line?) Why would anyone do that?  And if they do do it, should it need a warning?

hjoliver (commented):
Couple of minor questions.

cylc/rose/utilities.py
wxtim: Yes - It's a suite of configurations for a workflow.

cylc/rose/utilities.py
wxtim:
```suggestion
```

cylc/rose/utilities.py
oliver-sanders: Yes, this is a "thing", in Rose ignored items still exist in the configuration, just with a different state so could still be functional in some strange way.

I don't think we need a warning here, you'd have to go out of your way to add an ignored config, if you do we can only assume you know what you're doing.

cylc/rose/entry_points.py
```diff
@@ -120,6 +122,10 @@ def get_rose_vars(srcdir=None, opts=None):
             raise NotARoseSuiteException()
         return config
 
+    # Check for definitely invalid defines
+    if opts and hasattr(opts, 'defines') and not cylc7_back_compat:
```
oliver-sanders, line 126: Why are we skipping this check in `cylc7_back_compat` mode?

wxtim: This is a check that didn't exist at Cylc 7. We shouldn't modify the behaviour in compatibility mode.

MetRonnie:
> Built atop #221

#221 has unresolved comments

cylc/rose/entry_points.py
oliver-sanders: I think this would have resulted in an error at Cylc 7 just as it has at Cylc 8.

The exact error would be different, but either way, it wouldn't work with either. The point of this PR was to catch this easily detectable error early so I don't think that back-compat mode should come into it?

oliver-sanders: @wxtim, I don't think the back-compat mode check makes any sense here:

* This is a CLI interface.
* Missing the `=` symbol will result in the option being ignored.

oliver-sanders (approved):

hjoliver (approved):

PR #229, opened 2023-06-01
Fix.template vars section

wxtim:
Does not close a separate issue;

### Issue

Reported by @james-bruten-mo: 

Rose stem assumes that the name of the templating section is `[jinja2:suite.rc]`. At Rose 2  `[template variables]` is preferred, though the older form is kept for back compatibility. If the new form is used in the config file rose stems use of the older form will cause one of cylc-roses checks on the config to fail.

**Check List**

- [x] ~Doc changes~ are not needed.


MetRonnie (approved):
Looks ok to me but I don't know how to test

PR #241, opened 2023-07-24
Bump dev version

MetRonnie:


wxtim:
I think master should be 1.4.0

wxtim:
I think that master should be 1.4.0, even if I think we may have a fair few 1.3.x PRs

MetRonnie:
Can always bump to 1.4.0.dev afterwards, immediately afterwards if you want, but we'll need this on at least one branch anyway

PR #242, opened 2023-07-25
pin to next dev release

wxtim:
Part of https://github.com/cylc/cylc-admin/issues/174


MetRonnie:
IMO no need to do this for a while. There aren't any 1.4.0 milestone PRs in the queue

PR #244, opened 2023-08-02
Fix flake8

MetRonnie:


wxtim (approved):
Legitimate, and a small change.

PR #250, opened 2023-08-24
Fix.rose stem name is empty string

wxtim:
Issue documented in code and PR.

Before `-s=name` would lead to project name being manually set to '' rather than using FCM to setup a project name. This is a correct from the point of view of optparse, but might seem surprising to users.

@dpmatthews  tagged as product owner/understand why the change.
@MetRonnie  as code review - IMO it's so trivial we shouldn't spend much time on it.

**Check List**

### DIFF
--- a/tests/unit/test_rose_stem_units.py
+++ b/tests/unit/test_rose_stem_units.py
@@ -316,7 +316,7 @@ def test__deduce_mirror():
 )
 def test_ascertain_project_if_name_supplied(
     get_StemRunner: Fixture,
-    capsys: Fixture,
+    capsys: pytest.CaptureFixture,
     item: str,
     expect: Tuple[str],
     stdout: str,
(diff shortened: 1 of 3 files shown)


tests/unit/test_rose_stem_units.py
```diff
@@ -291,3 +295,45 @@ def test__deduce_mirror():
...
+            # Malformed project=url
+            '=foo',
+            ('', 'foo'),
+            None,
+        ),
+    )
+)
+def test_ascertain_project_if_name_supplied(
+    get_StemRunner: Fixture,
+    capsys: Fixture,
```
MetRonnie, line 319:
```suggestion
    capsys: pytest.CaptureFixture,
```

MetRonnie (approved):
LGTM, haven't tested

PR #264, opened 2023-10-25
GH Actions: use branch sync workflow

MetRonnie:


.github/workflows/branch_sync.yml
```diff
@@ -0,0 +1,20 @@
+name: Sync PR
+
+on:
+  push:
```
wxtim, line 4: Worth having a manual trigger too, if only for testing purposes?

wxtim (approved):
Yay!

PR #265, opened 2023-11-01
Rejig

oliver-sanders:
Bit of TLC in preparation for the addition of a new rose-stem interface which needs to be bodged into the middle of an existing function.

Main changes:

* refactor: move utilities out of the entry_points module    
  * Move utils out of the entry points module and move the rose-stem entry    
    point in.    
  * Move fileinstall into its own module.    
  * Sort imports.   
* entry_points: remove type uncertainty    
  * Entry points accept missing or None values for arguments, however,    
    this can never happen in practice.    
  * Set the appropriate type hints and remove code paths which could not    
    get executed.
  * Cascade this to simplify code further down the call chain

Minor changes:

* entry_points: remove unused results return value    
* utils: split the loading and processing of the config    
  * A future rose stem interface will need to inject itself    
    between config loading and processing.    
  * So colocate the plugin processing logic into one function.    
* fileinstall: remove unused "srcdir" argument    
* utils: remove unused "opts" arg in "rose_config_exists"    
* utils: make environ arg optional    
* tests: fix accidentally skipped tests    
  * `['a'].sort()` returns `None` not `['a']`    
  * So `assert a.sort() == b.sort()` will always be `True`.    
* tests: fix os.environ test interaction    
* tests: amend docstrings                                                                                                         
  * Amend false docstrings.    
  * Rename a couple of modules.    

**Check List**


cylc/rose/entry_points.py
```diff
@@ -13,320 +13,74 @@
...
-def record_cylc_install_options(
-    rundir=None,
-    opts=None,
-    srcdir=None,
-):
-    """Create/modify files recording Cylc install config options.
+def pre_configure(srcdir: Path, opts: 'Values') -> dict:
+    """Run before the Cylc configuration is read."""
+    if not srcdir:
+        # not sure how this could happen
```
wxtim, line 38: I'm not either - why did you add it?

oliver-sanders: This case was handled by the code before, I didn't get around to removing it.

MetRonnie: If `srcdir` is really a `Path`, then it can never be falsy

cylc/rose/utilities.py
```diff
@@ -451,9 +482,9 @@ def get_cli_opts_node(opts=None, srcdir=None):
         {'value': 'BAZ', 'state': '', 'comments': []}
     """
     # Unpack info we want from opts:
-    opt_conf_keys = []
-    defines = []
-    rose_template_vars = []
+    opt_conf_keys: list = []
+    defines: list = []
+    rose_template_vars: list = []
```
MetRonnie, line 487: Can it be more specific? `List[str]`?

oliver-sanders: Don't know the types, added the `list` placeholder purely to allow the function to by type checked in the first place.

MetRonnie (approved):

wxtim (approved):

PR #269, opened 2023-12-04
Repopulate cached GlobalConfig object if CYLC_SYMLINKS exported to env.

### LINKED ISSUE #237: Environment variables set in rose-suite.conf should apply when the global config is loaded
### Problem

Some users want to be able to choose a different symlink dirs setup per workflow, see https://github.com/cylc/cylc-flow/issues/5418.
This can be done via the use of environment variables and Jinja2 in the global config, e.g.:
```
#!jinja2
[install]
    [[symlink dirs]]
        [[[hpc]]]
{% if environ["CYLC_SYMLINKS"] | default("x") == "A" %}
            run = $PROJECT_A
{% elif environ["CYLC_SYMLINKS"] | default("x") == "B" %}
            run = $PROJECT_B
{% else %}
            run = $DATADIR
{% endif %}
```

Where appropriate (depending on the use case) this could be configured at the site level or the user level.

In order for this to work you need to be able to configure an environment variable to be set both for `cylc install` (for localhost) and for `cylc play` (for remote hosts). It is already possible to define environment variables in `rose-suite.conf` which alter the scheduler environment:
https://metomi.github.io/rose/doc/html/api/configuration/suite.html#rose:conf:rose-suite.conf[env]
However, at present these do not apply when the global config is loaded.

### Proposed Solution

Ensure that environment variables set in `rose-suite.conf` apply when the global config is loaded (or reloaded if necessary).

wxtim:
Closes #237 

This is now ready for review but needs testing against https://github.com/cylc/cylc-flow/pull/5839.

Ensure that the cached globalconfig object is reloaded after the export of `CYLC_SYMLINKS` variable. This means that using the `CYLC_SYMLINKS` variable allows users to specify installation symlink locations in the `rose-suite.conf`.

n.b. I was surprised to find that running `glbl_cfg()._DEFAULT = None` did not clear the cached config in a way that would force the reload in Cylc, but I think just reloading the config in Cylc Rose makes at least as much sense.

**Check List**

- [x] Docs: I have updated the plugin documentation.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,14 @@ creating a new release entry be sure to copy & paste the span tag with the
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 
+
+## __cylc-rose-1.4.0 (<span actions:bind='release-date'>Upcoming</span>)__
+
+### Features
+
+[#269](https://github.com/cylc/cylc-rose/pull/269) - Allow environment variables
+set in ``rose-suite.conf`` to be used when parsing ``global.cylc``.
+
 ## __cylc-rose-1.3.1 (<span actions:bind='release-date'>Released 2023-10-24</span>)__
 
 ### Fixes
--- a/cylc/rose/__init__.py
+++ b/cylc/rose/__init__.py
@@ -70,11 +70,37 @@ The following sections are permitted in the ``rose-suite.conf`` files:
    for ease of porting Cylc 7 workflows.
 
 
+The ``global.cylc`` file
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Cylc Rose Plugin forces the reloading of the ``global.cylc`` file
+to allow environment variables set by Rose to change the global configuration.
+
+For example you could use ``CYLC_SYMLINKS`` as a variable to control
+the behaviour of ``cylc install``:
+
+   .. code-block:: cylc
+
+      #!jinja2
+      # part of a global.cylc file
+      [install]
+         [[symlink dirs]]
+            [[[hpc]]]
+      {% if environ["CYLC_SYMLINKS"] | default("x") == "A" %}
+               run = $LOCATION_A
+      {% elif environ["CYLC_SYMLINKS"] | default("x") == "B" %}
+               run = $LOCATION_B
+      {% else %}
+               run = $LOCATION_C
+      {% endif %}
+
+
+
 Special Variables
 -----------------
 
 The Cylc Rose plugin provides two environment/template variables
-to the Cylc scheduler:
+to the Cylc scheduler.
 
 ``ROSE_ORIG_HOST``
    Cylc commands (such as ``cylc install``, ``cylc validate`` and
@@ -110,6 +136,7 @@ to the Cylc scheduler:
 
       ``CYLC_VERSION`` will be removed from your configuration by the
       Cylc-Rose plugin, as it is now set by Cylc.
+
 
 Additional CLI options
 ----------------------
--- a/cylc/rose/utilities.py
+++ b/cylc/rose/utilities.py
@@ -27,6 +27,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
 from cylc.flow import LOG
 from cylc.flow.exceptions import CylcError
 from cylc.flow.flags import cylc7_back_compat
+from cylc.flow.cfgspec.glbl_cfg import glbl_cfg
 from cylc.flow.hostuserutil import get_host
 from metomi.isodatetime.datetimeoper import DateTimeOperator
 from metomi.rose import __version__ as ROSE_VERSION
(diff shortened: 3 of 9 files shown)


CHANGES.md
```diff
@@ -6,6 +6,14 @@ creating a new release entry be sure to copy & paste the span tag with the
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 
+
+## __cylc-rose-1.4.0 (<span actions:bind='release-date'>Upcoming</span>)__
+
+### Features
+
+[#269](://github.com/cylc/cylc-rose/pull/269) - Allow Environment variables
```
MetRonnie, line 14:
```suggestion
[#269](https://github.com/cylc/cylc-rose/pull/269) - Allow environment variables
```

cylc/rose/__init__.py
MetRonnie, line 1: I think Oliver has said there is no special handling for `CYLC_SYMLINKS`?

wxtim:
Now failing identically to [master nightly tests](https://github.com/cylc/cylc-rose/actions/runs/7381378751/job/20079851194).

oliver-sanders (approved):
LGTM

tests/functional/test_utils.py
```diff
@@ -36,3 +36,84 @@ def test_basic(tmp_path):
...
+        '#!jinja2\n'
+        '[install]\n'
+        '    [[symlink dirs]]\n'
+        '        [[[localhost]]]\n'
+        '{% set cylc_symlinks = environ.get(\'CYLC_SYMLINKS\', None) %}\n'
+        '{% if cylc_symlinks == "foo" %}\n'
+        f'log = {str(tmp_path)}/foo\n'
+        '{% else %}\n'
+        f'log = {str(tmp_path)}/bar\n'
+        '{% endif %}\n'
```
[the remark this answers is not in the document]
wxtim: I agree, but escaping the Jinja2 would be a pain, where here I just make the lines with Jinja2 non-f-strings.

cylc/rose/__init__.py
```diff
@@ -70,11 +70,37 @@
...
+      [install]
+         [[symlink dirs]]
+            [[[hpc]]]
+      {% if environ["CYLC_SYMLINKS"] | default("x") == "A" %}
+               run = $LOCATION_A
+      {% elif environ["CYLC_SYMLINKS"] | default("x") == "B" %}
+               run = $LOCATION_B
+      {% else %}
+               run = $LOCATION_C
+      {% endif %}
```
MetRonnie, line 95:
```suggestion
.. code-block:: cylc

   #!jinja2
   # part of a global.cylc file
   [install]
      [[symlink dirs]]
         [[[hpc]]]
   {% if environ["CYLC_SYMLINKS"] | default("x") == "A" %}
            run = $LOCATION_A
   {% elif environ["CYLC_SYMLINKS"] | default("x") == "B" %}
            run = $LOCATION_B
   {% else %}
            run = $LOCATION_C
   {% endif %}
```

cylc/rose/__init__.py
```diff
@@ -70,11 +70,37 @@
...
+
+
+
 Special Variables
 -----------------
 
 The Cylc Rose plugin provides two environment/template variables
```
MetRonnie, line 102: (unrelated to your changes, but just spotted)
```suggestion
The Cylc Rose plugin provides three environment/template variables
```

tests/conftest.py
```diff
@@ -119,16 +131,21 @@ def _inner(srcpath, args=None):
         """
         options = Options(install_gop(), args)()
         output = SimpleNamespace()
+        if not options.workflow_name:
+            options.workflow_name = workflow_name
+        if not args or args and not args.get('no_run_name', ''):
```
MetRonnie, line 136:
```suggestion
        if not args or not args.get('no_run_name', ''):
```

tests/functional/test_utils.py
```diff
@@ -36,3 +36,92 @@ def test_basic(tmp_path):
...
+            [[x]]
+    """)
+
+    # Validate the config:
+    output = cylc_validate_cli(tmp_path)
+    assert output.ret == 0
+
+    # CYLC_SYMLINKS == None the first time the global.cylc
+    # is loaded and "Foo" the second time.
+    assert output.logging == '"Foo"'
```
MetRonnie, line 75: Running locally I get
```
        # CYLC_SYMLINKS == None the first time the global.cylc
        # is loaded and "Foo" the second time.
>       assert output.logging == '"Foo"'
E       assert 'None\n"Foo"' == '"Foo"'
E         + None
E           "Foo"

tests/functional/test_utils.py:75: AssertionError
```

wxtim:
@MetRonnie  - this is failing in the same way as master.

MetRonnie (approved):

cylc/rose/utilities.py
```diff
@@ -869,6 +870,13 @@ def export_environment(environment: Dict[str, str]) -> None:
     for key, val in environment.items():
         os.environ[key] = val
 
+    # If env vars have been set we want to force reload
+    # the global config so that the value of this vars
+    # can be used by Jinja2 in the global config.
+    # https://github.com/cylc/cylc-rose/issues/237
+    if environment:
```
oliver-sanders, line 877: Note, the environment is always populated as a result of automatic Rose variables.

PR #271, opened 2023-12-04
GH Actions: separate lint vs test jobs

MetRonnie:
`pip install` is currently failing on Python 3.7 due to conflicting dependencies related to flake8.

I have separated the linting and testing dependencies and separated them into different jobs on GH Actions

### DIFF
--- a/.github/workflows/branch_sync.yml
+++ b/.github/workflows/branch_sync.yml
@@ -1,0 +1,20 @@
+name: Sync PR
+
+on:
+  push:
+    branches:
+      - '*.*.x'
+  schedule:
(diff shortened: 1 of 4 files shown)


MetRonnie:
@wxtim any idea about the test failures?

oliver-sanders (approved):

wxtim:
> @wxtim any idea about the test failures?

My immediate suspicion is that it's the xdist settings: Those rose stem tests come in groups and if they aren't kept with those groups that may cause failure. 

I'm going to have a play.

wxtim (approved):

PR #289, opened 2024-01-19
Mock user config without creating a config file.

wxtim:
Prevents caching of the rose config from making some rose stem tests flaky.
As a side benefit it's a little quicker.

How to demonstrate the problem:

```console
test1="tests/functional/test_rose_fileinstall.py::test_rose_fileinstall_run"
test2="tests/functional/test_rose_stem.py::TestWithConfig"

pytest $test1 $test2
# ... fail
pytest $test2 $test1
# ... pass
```

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] This is a fix to testing, and does not need a change log entry or documentation changes.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

PR #291, opened 2024-01-25
Mock only a single method of the rose config object.

wxtim:
Small change to #289 to avoid the use of mock, and to change the Rose config object as little as possible.

### DIFF
--- a/tests/functional/test_rose_stem.py
+++ b/tests/functional/test_rose_stem.py
@@ -120,7 +120,9 @@ def mock_global_cfg(monkeymodule):
         """Get the ResourceLocator.default and patch its get_conf method
         """
         obj = ResourceLocator.default()
-        obj.get_conf = lambda: ConfigLoader().load(StringIO(conf))
+        monkeymodule.setattr(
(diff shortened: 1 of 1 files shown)


tests/functional/test_rose_stem.py
```diff
@@ -118,13 +117,12 @@ def mock_global_cfg(monkeymodule):
...
     def _inner(target, conf):
-        """Mock a config object with a config node."""
-        node = ConfigLoader().load(StringIO(conf))
-
-        # Create a fake class imitating ConfigTree with .get_conf method:
-        config = MagicMock(spec=['get_conf'], get_conf=lambda: node)
+        """Get the ResourceLocator.default and patch its get_conf method
+        """
+        obj = ResourceLocator.default()
+        obj.get_conf = lambda: ConfigLoader().load(StringIO(conf))
```
MetRonnie, line 123: I think this will leave the patching in place after the test has run?

It would be safer to monkeypatch it
```suggestion
        monkeymodule.setattr(
            obj, 'get_conf', lambda: ConfigLoader().load(StringIO(conf))
        )
```

oliver-sanders (approved):

PR #302, opened 2024-03-18
platforms: use processed workflow config

oliver-sanders:
* When retrieving the platform used by a task, use the processed workflow config rather than processing the config from scratch.
* This avoids the need to access the DB to extract template variables (workflow params) which could cause public DB locking issues.

**Check List**

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,13 @@ creating a new release entry be sure to copy & paste the span tag with the
 updated. Only the first match gets replaced, so it's fine to leave the old
 ones in. -->
 
+## __cylc-rose-1.3.3 (<span actions:bind='release-date'>Awaiting Release</span>)__
+
+[#300](https://github.com/cylc/cylc-rose/pull/300) -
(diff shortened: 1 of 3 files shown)


oliver-sanders:
Kicked off Rose tests against this branch: https://github.com/metomi/rose/actions/runs/8325352084

wxtim (approved):
- [x] Read code change
- [x] Checked where code is likely to touch the database using `./t/rose-task-run/04-run-path-empty`
- [x] Run test manually.

MetRonnie (approved):

MetRonnie (approved):
Re-ran tests

hjoliver:
I think this will fix a bug we've just run into at NIWA. We have a custom Python module that determines Slurm accounting code from a project code in the environment. The `fcm_make` Rose app is failing when it parses the `flow.cylc` to get platforms info, because the PYTHONPATH to the project code module isn't getting through to that process (even though it works fine for Cylc at at scheduler start-up).

oliver-sanders:
That makes sense.

PR #308, opened 2024-04-08
clean the options with information for cylc rose after dumping thier …

wxtim:
…content in rose-suite-cylc-install.conf

Partner of https://github.com/cylc/cylc-flow/pull/6060

**Check List**

### DIFF
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -71,6 +71,12 @@ def post_install(srcdir: Path, rundir: str, opts: 'Values') -> bool:
     if config_node:
         dump_rose_log(rundir=_rundir, node=config_node)
 
+    # Clear opts before returning; they have now been stored in
+    # rose-suite-cylc-install.conf
+    opts.rose_template_vars = []
+    opts.defines = []
+    opts.opt_conf_keys = []
+
     return True
 
 
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -73,6 +73,7 @@ def post_install(srcdir: Path, rundir: str, opts: 'Values') -> bool:
 
     # Clear opts before returning; they have now been stored in
     # rose-suite-cylc-install.conf
+    # https://github.com/cylc/cylc-flow/issues/6058
     opts.rose_template_vars = []
     opts.defines = []
     opts.opt_conf_keys = []
--- a/tests/unit/test_functional_post_install.py
+++ b/tests/unit/test_functional_post_install.py
@@ -355,3 +355,20 @@ def test_cylc_no_rose(tmp_path):
 def test_copy_config_file_fails(tmp_path):
     """It fails if source not a rose suite."""
     copy_config_file(tmp_path, tmp_path)
+
+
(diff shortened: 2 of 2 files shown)


tests/unit/test_functional_post_install.py
```diff
@@ -355,3 +355,20 @@ def test_cylc_no_rose(tmp_path):
...
+    monkeypatch.setattr(f'{path}copy_config_file', lambda *_, **__: False)
+    monkeypatch.setattr(f'{path}rose_config_exists', lambda *_, **__: True)
+
+    opts = SimpleNamespace(
+        rose_template_vars=['gee', 'whiz'],
+        defines=['foo', 'bar'],
+        opt_conf_keys=['thomas', 'percy'])
+
+    post_install('', '', opts)
+    assert any(opts.__dict__.values()) is False
```
oliver-sanders, line 374: For a functional test, how about something installs and then reinstalls a workflow using the same `opts` object e.g:

```python
def test_something(cylc_install_cl):
    opts = SimpleNamespace(
        rose_template_vars=['gee', 'whiz'],
        defines=['foo', 'bar'],
        opt_conf_keys=['thomas', 'percy']
    )

    # install
    run_dir = cylc_install_cli(..., opts)

    # check installed rose config
    RoseConfigTree(run_dir).get(('template variables', 'gee')) == 'whatever'

    # reinstall using same opts object (just like cylc-flow does)
    run_dir = cylc_install_cli(..., opts)

     # check the installed rose config again
    RoseConfigTree(run_dir).get(('template variables', 'gee')) == 'whatever'
```

cylc/rose/entry_points.py
```diff
@@ -71,6 +71,12 @@ def post_install(srcdir: Path, rundir: str, opts: 'Values') -> bool:
     if config_node:
         dump_rose_log(rundir=_rundir, node=config_node)
 
+    # Clear opts before returning; they have now been stored in
+    # rose-suite-cylc-install.conf
```
oliver-sanders, line 75: Please link to issue.

wxtim:
```suggestion
    # Clear opts before returning; they have now been stored in
    # rose-suite-cylc-install.conf
    # https://github.com/cylc/cylc-flow/issues/6058
```

tests/unit/test_functional_post_install.py
wxtim: >     # reinstall using same opts object (just like cylc-flow does)

Sorry, but Cylc Flow doesn't reinstall with the same opts object as Cylc install - do you mean play?

wxtim: I don't understand how this is supposed to demonstate the broken state before my changes. `cylc reinstall` interacts with the Opts and the source directory (the validate step in VR does the interaction with the run dir). In this bug the problem is the interaction between the opts and the run directory when `cylc play` is run. I'm happy to do _that_ as a functional test.

wxtim: I don't think that this makes sense: the problem was with the content of the run directory interacting with `cylc play`, not the source directory. I'll translate my manual test into one of these.

PR #310, opened 2024-04-12
Test `-S`/`--rose-template-variable`

wxtim:
Tests for https://github.com/cylc/cylc-flow/pull/6065

Full explanation in that PR.

**Check List**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,6 +18,7 @@
 
 import pytest
 from types import SimpleNamespace
+from uuid import uuid4
 
 from cylc.flow import __version__ as CYLC_VERSION
 from cylc.flow.option_parsers import Options
@@ -36,6 +37,11 @@ from cylc.flow.scripts.reinstall import (
     reinstall_cli as cylc_reinstall,
     get_option_parser as reinstall_gop
 )
+
+
+def test_workflow_name():
+    """Return a UUID to use as a test name"""
+    return f'cylc-rose-test-{str(uuid4())[:8]}'
 
 
 @pytest.fixture(scope='module')
@@ -128,8 +134,14 @@ def _cylc_install_cli(capsys, caplog):
             srcpath:
             args: Dictionary of arguments.
         """
+        if not args or not args.get('workflow_name', ''):
+            id_ = test_workflow_name()
+            args = {'workflow_name': id_}
+
+        args.update({'no_run_name': True})
         options = Options(install_gop(), args)()
         output = SimpleNamespace()
+        output.id = args['workflow_name']
 
         try:
             cylc_install(options, str(srcpath))
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -16,8 +16,13 @@
 """Functional tests for top-level function record_cylc_install_options and
 """
 
+from pathlib import Path
 import pytest
+from subprocess import run
+from shlex import split
+from time import sleep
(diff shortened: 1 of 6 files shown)


tests/functional/test_reload.py
```diff
@@ -0,0 +1,148 @@
...
+    assert 'cylc message -- CLIreinstall' in config_log.read_text()
+
+    purge_workflow(install_results.id)
+
+
+def test_restart_overrides(
+    cylc_install_cli, cylc_reinstall_cli, tmp_path, file_poll,
+    provide_template_vars_workflow, purge_workflow, cylc_stop
+):
+    """When we restart a workflow, the play CLI options are honoured.
```
wxtim, line 99: @oliver-sanders  - I've included this to demonstrate that I think that the clearance of the arguments shouldn't happen if the reload is one that comes with a restart since restart can have options set, unlike reload, and that if they are set then the user would expect the CLI variables to override those in the rose files, even if we _have_ just reinstalled too.

tests/conftest.py
```diff
@@ -197,3 +218,39 @@ def cylc_validate_cli(capsys, caplog):
...
+                f'file {fpath} not found after {timeout} seconds')
+    return _inner
+
+
+@pytest.fixture
+def cylc_stop(file_poll):
+    """Cylc stop & check the contact file"""
+    def _inner(id_):
+        run(split(f'cylc stop --now --now {id_}'))
+        contactfile = Path.home() / f'cylc-run{id_}.service/contact'
```
MetRonnie, line 245: Shouldn't this be?
```suggestion
        contactfile = Path.home() / f'cylc-run/{id_}/.service/contact'
```
Or better yet
```python
from cylc.flow.workflow_files import get_contact_file_path
```

tests/functional/test_reload.py
```diff
@@ -0,0 +1,148 @@
...
+    (tmp_path / 'flow.cylc').write_text(
+        '#!jinja2\n'
+        '[scheduling]\n'
+        '    [[graph]]\n'
+        '       R1 = foo\n'
+        '[runtime]\n'
+        '    [[foo]]\n'
+        '        script = cylc message -- {{var}}')
+    (tmp_path / 'rose-suite.conf').write_text(
+        '[template variables]\nvar="rose-suite.conf"')
```
MetRonnie, line 46: N.B. I think it would be easier and cleaner to use a triple quoted string and `textwrap.dedent()`

tests/functional/test_reload.py
```diff
@@ -0,0 +1,148 @@
...
+1. Validates a workflow.
+2. Installs a workflow with some opts set using -O and
+   ROSE_SUITE_OPT_CONF_KEYS.
+3. Re-install workflow.
+4. After modifying the source ``rose-suite.conf``, re-install the flow again.
+
+At each step it checks the contents of
+- ~/cylc-run/temporary-id/rose-suite.conf
+- ~/cylc-run/temporary-id/opt/rose-suite-cylc-install.conf
+"""
```
MetRonnie, line 28: This does not seem accurate. Copied from another test?

tests/functional/test_reload.py
MetRonnie, line 1: What about `cylc vr`?

wxtim:
Closing becuase after an offline discussion it was decided to approach a totally different solution.

PR #312, opened 2024-04-16
Delete opts after post install

wxtim:
Closes https://github.com/cylc/cylc-flow/issues/5968 alongside a https://github.com/cylc/cylc-flow/pull/6068.

Additionally closes https://github.com/cylc/cylc-flow/issues/6058, which was caused by an earlier incomplete solution https://github.com/cylc/cylc-flow/pull/5996

Rose variables should be removed from the options object after post install plugin to prevent them being used by any downstream Cylc Scripts.

**Check List**

### DIFF
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -56,7 +56,7 @@ class NotARoseSuiteException(Exception):
 
 
 def pre_configure(srcdir, opts):
-    srcdir, rundir = paths_to_pathlib([srcdir, rundir])
+    srcdir, = paths_to_pathlib([srcdir,])
     return get_rose_vars(srcdir=srcdir, opts=opts)
 
 
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -56,7 +56,7 @@ class NotARoseSuiteException(Exception):
 
 
 def pre_configure(srcdir, opts):
-    srcdir, = paths_to_pathlib([srcdir,])
+    srcdir = Path(srcdir)
     return get_rose_vars(srcdir=srcdir, opts=opts)
 
 
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -75,7 +75,9 @@ def post_install(srcdir, opts, rundir):
     if results['fileinstall']:
         dump_rose_log(rundir=rundir, node=results['fileinstall'])
 
-    # Having dumped the config we clear rose options:
+    # Having dumped the config we clear rose options
+    # as they do not apply after this.
+    # see https://github.com/cylc/cylc-rose/pull/312
     opts.rose_template_vars = []
     opts.opt_conf_keys = []
     opts.defines = []
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -54,12 +54,12 @@ class NotARoseSuiteException(Exception):
         return msg
 
 
-def pre_configure(srcdir=None, opts=None, rundir=None):
+def pre_configure(srcdir, opts, rundir=None):
     srcdir, rundir = paths_to_pathlib([srcdir, rundir])
     return get_rose_vars(srcdir=srcdir, opts=opts)
 
 
-def post_install(srcdir=None, opts=None, rundir=None):
+def post_install(srcdir, opts, rundir):
     if not rose_config_exists(srcdir, opts):
         return False
     srcdir, rundir = paths_to_pathlib([srcdir, rundir])
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -55,7 +55,7 @@ class NotARoseSuiteException(Exception):
         return msg
 
 
-def pre_configure(srcdir, opts, rundir=None):
+def pre_configure(srcdir, opts):
     srcdir, rundir = paths_to_pathlib([srcdir, rundir])
     return get_rose_vars(srcdir=srcdir, opts=opts)
 
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -75,6 +75,11 @@ def post_install(srcdir=None, opts=None, rundir=None):
     if results['fileinstall']:
         dump_rose_log(rundir=rundir, node=results['fileinstall'])
 
+    # Having dumped the config we clear rose options:
+    opts.rose_template_vars = []
+    opts.opt_conf_keys = []
+    opts.defines = []
+
     return results
 
 
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -27,6 +27,7 @@ from pathlib import Path
 from metomi.rose.config import ConfigLoader, ConfigDumper
 from cylc.rose.utilities import (
     ROSE_ORIG_HOST_INSTALLED_OVERRIDE_STRING,
+    ROSE_SUITE_OPT_CONF_KEYS,
     deprecation_warnings,
     dump_rose_log,
     get_rose_vars_from_config_node,
@@ -81,6 +82,7 @@ def post_install(srcdir, opts, rundir):
     opts.rose_template_vars = []
     opts.opt_conf_keys = []
     opts.defines = []
+    os.unsetenv(ROSE_SUITE_OPT_CONF_KEYS)
 
     return results
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,7 +56,7 @@ python_requires = >=3.7
 include_package_data = True
 install_requires =
     metomi-rose>=2.1.0, <2.3.0
-    cylc-flow>=8.2.6, <8.3
+    cylc-flow>8.2.5, <8.3
     metomi-isodatetime
     ansimarkup
     jinja2
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,7 +56,7 @@ python_requires = >=3.7
 include_package_data = True
 install_requires =
     metomi-rose>=2.1.0, <2.3.0
-    cylc-flow>=8.2.6
+    cylc-flow>=8.2.6, <8.3
     metomi-isodatetime
     ansimarkup
     jinja2
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,7 +56,7 @@ python_requires = >=3.7
 include_package_data = True
 install_requires =
     metomi-rose>=2.1.0, <2.3.0
-    cylc-flow==8.2.*
+    cylc-flow>=8.2.6
     metomi-isodatetime
     ansimarkup
     jinja2
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -222,16 +222,11 @@ def file_poll():
     def _inner(
         fpath: "Path", timeout: int = 5, inverse: bool = False
     ):
-        if inverse:
-            def check(f):
-                return not f.exists()
-        else:
-            def check(f):
-                return f.exists()
-
         timeout_func(
-            partial(check, fpath),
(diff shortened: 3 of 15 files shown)


tests/unit/test_functional_post_install.py
```diff
@@ -353,10 +353,3 @@ def broken():
     (tmp_path / 'rose-suite.conf').touch()
     with pytest.raises(FileNotFoundError):
         rose_fileinstall(srcdir=tmp_path, rundir=tmp_path)
-
-
-def test_cylc_no_rose(tmp_path):
```
wxtim, line 358: Precisely duplicated `test_no_rose_suite_conf_in_devdir` at line 60.

setup.cfg
```diff
@@ -56,7 +56,7 @@ python_requires = >=3.7
 include_package_data = True
 install_requires =
     metomi-rose>=2.1.0, <2.3.0
-    cylc-flow==8.2.*
+    cylc-flow>=8.2.6
```
wxtim, line 59: This will may no longer work with lower versions of Cylc.

cylc/rose/entry_points.py
```diff
@@ -75,6 +75,11 @@ def post_install(srcdir=None, opts=None, rundir=None):
     if results['fileinstall']:
         dump_rose_log(rundir=rundir, node=results['fileinstall'])
 
+    # Having dumped the config we clear rose options:
+    opts.rose_template_vars = []
+    opts.opt_conf_keys = []
+    opts.defines = []
```
MetRonnie, line 81: Pylance points out `opts` could be `None`.

However I don't think any of these arguments should have a default in the function definition
```diff
-def post_install(srcdir=None, opts=None, rundir=None):
+def post_install(srcdir, opts, rundir):
```

tests/conftest.py
```diff
@@ -197,3 +217,50 @@ def cylc_validate_cli(capsys, caplog):
...
+        )
+    return _inner
+
+
+@pytest.fixture
+def run_ok():
+    """Run a bash script.
+    Fail if it fails and return its output.
+    """
+    def _inner(script):
```
MetRonnie, line 252: VSCode/pyright does now include type info in intellisense for pytest fixtures, so would be useful to include
```suggestion
    def _inner(script: str):
```

tests/functional/test_reinstall.py
```diff
@@ -280,3 +282,58 @@ def test_reinstall_workflow(tmp_path):
...
+    (tmp_path / 'flow.cylc').write_text(
+        '#!jinja2\n'
+        '[scheduling]\n'
+        '    [[graph]]\n'
+        '       R1 = foo\n'
+        '[runtime]\n'
+        '    [[foo]]\n'
+        '        script = cylc message -- {{var}}')
+    (tmp_path / 'rose-suite.conf').write_text(
+        '[template variables]\nvar="rose-suite.conf"')
```
MetRonnie, line 313: (Would be clearer IMO to use triple quoted strings and `textwrap.dedent()`)

cylc/rose/entry_points.py
oliver-sanders: The `post_install` interface hasn't always looked this way, probably hangover:

https://github.com/cylc/cylc-flow/blob/54959200e337c66df4896c066f042ffd2745f0d1/cylc/flow/scripts/install.py#L107-L109

We can drop the `=None` stuff (tests depending) with confidence because of Cylc version pinning. Just check the oldest cylc-flow version in range to make sure it's still compatible.

setup.cfg
oliver-sanders: You will need to reinstate the upper pin though:

```suggestion
    cylc-flow>=8.2.6, <8.3
```

tests/conftest.py
```diff
@@ -128,8 +142,14 @@ def _inner(srcpath, args=None):
...
             args: Dictionary of arguments.
         """
+        if not args or not args.get('workflow_name', ''):
+            id_ = test_workflow_name()
+            args = {'workflow_name': id_}
+
+        args.update({'no_run_name': True})
         options = Options(install_gop(), args)()
         output = SimpleNamespace()
+        output.id = args['workflow_name']
```
oliver-sanders, line 152: FYI, this is going to conflict when merged into master because all the `output.ret`, `output.exc`, `output.out`, etc stuff was stripped in favour of native pytest functionalities.

I think what you've added here mirrors the behaviour on master so should be easy to reconcile.

https://github.com/cylc/cylc-rose/blob/ed485a84b0f375d4dee2b7df7676d92712f0f905/tests/conftest.py#L175-L205

cylc/rose/entry_points.py
```diff
@@ -75,6 +75,11 @@ def post_install(srcdir=None, opts=None, rundir=None):
     if results['fileinstall']:
         dump_rose_log(rundir=rundir, node=results['fileinstall'])
 
+    # Having dumped the config we clear rose options:
```
oliver-sanders, line 78:
```suggestion
    # Having dumped the config we clear rose options:
    # see https://github.com/cylc/cylc-rose/pull/312
```

MetRonnie: Or better yet
```suggestion
    # Having dumped the config we clear rose options
    # as they do not apply after this.
    # see https://github.com/cylc/cylc-rose/pull/312
```

cylc/rose/entry_points.py
```diff
@@ -54,12 +54,12 @@ def __str__(self):
...
 
 
-def pre_configure(srcdir=None, opts=None, rundir=None):
+def pre_configure(srcdir, opts, rundir=None):
     srcdir, rundir = paths_to_pathlib([srcdir, rundir])
     return get_rose_vars(srcdir=srcdir, opts=opts)
 
 
-def post_install(srcdir=None, opts=None, rundir=None):
+def post_install(srcdir, opts, rundir):
```
oliver-sanders, line 62: Hmmmm:

https://github.com/cylc/cylc-flow/blob/40139b13ca15b1f1e21a2da6c81d9074dc8c5124/cylc/flow/scripts/reinstall.py#L291

MetRonnie: It's ok actually: https://github.com/cylc/cylc-flow/blob/7cf4bf1eef595c53571e00900ef1e8d956fac5b6/cylc/flow/scripts/reinstall.py#L352-L362

The `post_install` function in `cylc/flow/scripts/reinstall.py` has a different order of args but calls the entry point with `srcdir, opts, rundir` in that order

setup.cfg
```diff
@@ -56,7 +56,7 @@ python_requires = >=3.7
 include_package_data = True
 install_requires =
     metomi-rose>=2.1.0, <2.3.0
-    cylc-flow==8.2.*
+    cylc-flow>=8.2.6, <8.3
```
MetRonnie, line 59: Try this so that 8.2.6.dev0 gets allowed
```suggestion
    cylc-flow>8.2.5, <8.3
```

cylc/rose/entry_points.py
```diff
@@ -75,6 +75,13 @@ def post_install(srcdir=None, opts=None, rundir=None):
     if results['fileinstall']:
         dump_rose_log(rundir=rundir, node=results['fileinstall'])
 
+    # Having dumped the config we clear rose options
+    # as they do not apply after this.
+    # see https://github.com/cylc/cylc-rose/pull/312
+    opts.rose_template_vars = []
+    opts.opt_conf_keys = []
+    opts.defines = []
```
oliver-sanders, line 83: There's one thing that we're not clearing, `ROSE_SUITE_OPT_CONF_KEYS`.

wxtim:
> @wxtim, tests are still failing I'm afraid.

I think that I've fixed it. I don't like the functional testing I've built, but because I don't anticipate there ever being a need for much functional testing. Cylc Rose should _never_ have needed these tests, and they only demonstrate that it doesn't have functionality it wasn't meant to have.

cylc/rose/entry_points.py
```diff
@@ -54,12 +55,12 @@ def __str__(self):
         return msg
 
 
-def pre_configure(srcdir=None, opts=None, rundir=None):
+def pre_configure(srcdir, opts, rundir=None):
```
MetRonnie, line 58: I have just spotted, the `pre_configure` entry point does not actually provide the `rundir` arg

tests/conftest.py
```diff
@@ -223,24 +219,39 @@ def mod_cylc_validate_cli(mod_capsys, mod_caplog):
...
-    def _inner(fpath: "Path", timeout: int = 5):
+    def _inner(
+        fpath: "Path", timeout: int = 5, inverse: bool = False
+    ):
+        if inverse:
+            def check(f):
+                not f.exists()
+        else:
+            def check(f):
+                f.exists()
```
MetRonnie, line 230: These are not returning anything!

MetRonnie:
> I think that I've fixed it.
> 
> I don't like the functional testing I've built, but because I don't anticipate there ever being a need for much functional testing I don't think it's worth building anything better.
> 
> Cylc Rose should _never_ have needed these tests, and they only demonstrate that it doesn't have functionality it wasn't meant to have.

You seem to have removed the tests you added. At the moment there are no tests added in this PR

tests/conftest.py
```diff
@@ -197,3 +213,65 @@ def cylc_validate_cli(capsys, caplog):
...
+
+        timeout_func(
+            partial(check, fpath),
+            f'file {fpath} not found after {timeout} seconds'
+        )
```
MetRonnie, line 235: FWIW, if there are meant to be tests relying on this
```suggestion
        timeout_func(
            lambda: fpath.exists() != inverse,
            f"file {fpath} {'still' if inverse else 'not'} found after "
            f"{timeout} seconds",
            timeout
        )
```

cylc/rose/entry_points.py
wxtim: No, but there's no reason someone else might not use the interface.

wxtim:
```suggestion
def pre_configure(srcdir, opts):
```

cylc/rose/entry_points.py
```diff
@@ -54,12 +55,12 @@ def __str__(self):
         return msg
 
 
-def pre_configure(srcdir=None, opts=None, rundir=None):
-    srcdir, rundir = paths_to_pathlib([srcdir, rundir])
+def pre_configure(srcdir, opts):
+    srcdir, = paths_to_pathlib([srcdir,])
```
MetRonnie, line 59:
```suggestion
    srcdir = Path(srcdir)
```

tests/conftest.py
```diff
@@ -197,3 +213,60 @@ def cylc_validate_cli(capsys, caplog):
...
+        timeout_func(
+            partial(run_ok, stop),
+            message=f'Not run after {timeout} seconds: {stop}')
+        file_poll(
+            Path.home() / 'cylc-run' / id_ / '.service/contact',
+            inverse=True,
+        )
+        timeout_func(
+            partial(run_ok, clean),
+            message=f'Not run after {timeout} seconds: {clean}')
```
MetRonnie, line 249: (Not passing the timeout on here)

```suggestion
            message=f'Not run after {timeout} seconds: {stop}',
            timeout
        )
        file_poll(
            Path.home() / 'cylc-run' / id_ / '.service/contact',
            inverse=True,
        )
        timeout_func(
            partial(run_ok, clean),
            message=f'Not run after {timeout} seconds: {clean}',
            timeout
        )
```

MetRonnie (approved):
Couple of small suggestions

MetRonnie (approved):

PR #268, opened 2023-11-27
tests/functional: simplify rose-stem tests

oliver-sanders:
* Change class scoped fixtures to module scoped. This avoids test classes and ensures fixtures are only initiated once.
* Merge fixtures which were only used once in with the test which used them.
* Remove parameterisations and fail testing code which weren't used by the tests.
* Remove monkeypatch.delenv where it wasn't needed (monkeypatch resets things by default).

**Check List**

- [x] I have read `CONTRIBUTING.md` and added my name as a Code Contributor.
- [x] Contains logically grouped changes (else tidy your branch by rebase).
- [x] Does not contain off-topic changes (use other PRs for other changes).
- [x] Applied any dependency changes to both `setup.cfg` (and `conda-environment.yml` if present).
- [x] Tests are included (or explain why tests are not needed).
- [x] `CHANGES.md` entry included if this is a change that can affect users
- [x] [Cylc-Doc](https://github.com/cylc/cylc-doc) pull request opened if required at cylc/cylc-doc/pull/XXXX.
- [x] If this is a bug fix, PR should be raised against the relevant `?.?.x` branch.

Issue #205, opened 2023-01-16
stem: re-implement as Cylc install plugin


oliver-sanders:
The `rose stem` command wraps `cylc install` which is a bit awkward.

* The bespoke command is an additional complication to working practices.
* Cylc commands like `cylc validate` and `cylc graph` may fail on the source due to missing template variables.
* It makes Stem seem like a system onto itself whereas it is really just setting a couple of template variables.

Consider re-implementing `rose stem` as a pure plugin so that the Stem working practices are the same as for any other Cylc workflow. This would require Stem workflows to use a flag file or load a Stem plugin in some way.

For example if we were to implement a template variable file for Cylc, a Stem workflow might look something like:

```python
# cylc_template_variable_file.py

from cylc.rose.stem import get_stem_variables

def get_template_variables():
    # this line makes a workflow a `rose stem` workflow
    return get_stem_variables()
```

Otherwise we could use a flag file or configuration something like:

```ini
# rose-suite.conf

[stem]  # activate rose-stem
groups=foo bar baz  # define defaults

[template variables]
OPT_1=True
OPT_2=False
```

PR #314, opened 2024-04-23
simplify rose stem tests

oliver-sanders:
I'm planning to work on rose-stem with #205 so trying to get to grips with the tests. This:

* Simplifies the tests.
* Links them to their original source (metomi/rose).
* *Attempts* to give them meaningful names/descriptions.

Changes outlined in the commit messages.

Supersedes #268

**Check List**

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -62,6 +62,14 @@ def event_loop():
     loop.close()
 
 
+@pytest.fixture(scope='module')
+def monkeymodule():
+    """Make monkeypatching available in a module scope."""
+    mpatch = pytest.MonkeyPatch()
+    yield mpatch
+    mpatch.undo()
+
+
 @pytest.fixture()
 def workflow_name():
     return 'cylc-rose-test-' + str(uuid4())[:8]
@@ -97,8 +105,7 @@ def mod_caplog(request):
 
 @pytest.fixture(scope='package', autouse=True)
 def set_cylc_version():
-    from _pytest.monkeypatch import MonkeyPatch
-    mpatch = MonkeyPatch()
+    mpatch = pytest.MonkeyPatch()
     yield mpatch.setenv('CYLC_VERSION', CYLC_VERSION)
     mpatch.undo()
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import asyncio
+from io import StringIO
 from pathlib import Path
 from shutil import rmtree
 from types import SimpleNamespace
@@ -32,6 +33,14 @@ from cylc.flow.scripts.reinstall import reinstall_cli as cylc_reinstall
 from cylc.flow.scripts.validate import run as cylc_validate
 from cylc.flow.scripts.validate import get_option_parser as validate_gop
 from cylc.flow.wallclock import get_current_time_string
+
+from metomi.rose.resource import ResourceLocator
+from metomi.rose.config import ConfigLoader
+
+from cylc.rose.stem import (
+    get_rose_stem_opts,
+    rose_stem as _rose_stem,
+)
 
 
 CYLC_RUN_DIR = Path(get_cylc_run_dir())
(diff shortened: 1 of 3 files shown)


tests/functional/test_rose_stem.py
```diff
@@ -18,60 +18,16 @@
...
-Tests for cylc.rose.stem
-========================
-
-These tests are based on the Rose 2019 tests for Rose Stem.
-
-Structure
----------
-
-#. ``setup_stem_repo`` is a module scoped fixture which creates a Rose Stem
-   repository which is used for all the tests.
```
oliver-sanders, line 31: These docs were broken by previous changes.

The relevant info should be in the docstrings of the fixtures / methods.

tests/functional/test_rose_stem.py
```diff
@@ -228,316 +164,261 @@ def setup_stem_repo(tmp_path_factory, monkeymodule, request):
...
-    else:
-        expected = expected.format(
-            workingcopy=suite_redirection['workingcopy'],
-            hostname=HOST
-        )
-        assert expected in suite_redirection['jobout_content']
+    template_vars = await rose_stem_runner(
+        rose_stem_opts,
+        # don't CD into the project directory first
+        cwd=Path.cwd(),
```
oliver-sanders, line 316: Small test fix, this test was being run in the project directory.

tests/functional/test_rose_stem.py
```diff
@@ -228,316 +164,261 @@ def setup_stem_repo(tmp_path_factory, monkeymodule, request):
...
-    yield await rose_stem_run_template(rose_stem_opts)
-
-
-@pytest.mark.parametrize(
-    'expected',
-    [
-        "run_ok",
-        (
-            "RUN_NAMES=[\'earl_grey\', \'milk\', \'sugar\', "
-            "\'spoon\', \'cup\', \'milk\']"
```
oliver-sanders, line 357: Normalized strings ala Black which says we should switch quote style in preference to escaping (as does pep8 I think).

MetRonnie (approved):
Looks good however I seem to consistently get one of the tests failing to clean up afterwards with

> ERROR tests/functional/test_rose_stem.py::test_source_in_a_subdirectory - OSError: [Errno 39] Directory not empty: 'config'

(not necessarily `test_source_in_a_subdirectory` every time)

MetRonnie:
Having said that, a lot more directories are left over than just the 4 I would expect from running the tests 4 times

image

oliver-sanders:
> Having said that, a lot more directories are left over than just the 4 I would expect from running the tests 4 times

I tried running `pytest tests/functional/test_rose_stem.py` until I got a test failure which happened on the 5th run.

I got two directories left behind:

```console
$ ls -1 ~/cylc-run
cylc-rose-stem-test-38e5e755
cylc-rose-stem-test-f7445a1f
```

Which is one more than I would have expected, one was from `test_picks_template_section`, the other from `test_automatic_options_multi` (the test that actually failed).

I'm not sure why a failure in one test caused the other to be left behind.

oliver-sanders:
Ok, went a bit further with these tests and unified the rose-stem fixture with the others in conftest.py.

It is now using the same logic as other fixtures for creating the run-directory and removing it afterwards. All rose-stem tests will go into the root cylc-rose-test directory and only the failed test will be left behind.

tests/conftest.py
```diff
@@ -62,6 +71,14 @@ def event_loop():
     loop.close()
 
 
+@pytest.fixture(scope='module')
+def monkeymodule():
+    """Make monkeypatching available in a module scope."""
+    mpatch = pytest.MonkeyPatch()
+    yield mpatch
+    mpatch.undo()
```
wxtim, line 79: Apparently this can now (Pytest ≤ 6.2) be a bit simpler https://github.com/pytest-dev/pytest/issues/363#issuecomment-1335631998

```suggestion
@pytest.fixture(scope='module')
def monkeymodule():
    """Make monkeypatching available in a module scope."""
    with pytest.MonkeyPatch.context() as mp:
        yield mp
```

tests/conftest.py
```diff
@@ -259,6 +275,80 @@ def mod_cylc_validate_cli(mod_capsys, mod_caplog):
     return _cylc_validate_cli(mod_capsys, mod_caplog)
 
 
+@pytest.fixture
+def rose_stem(test_dir, monkeypatch, request):
+    """The Rose Stem command.
+
+    Wraps the "rose_stem" async function for use in tests.
+
+    Cleans up afterwards if the test was successfull.
```
wxtim, line 284:
```suggestion
    Cleans up afterwards if the test was successful.
```

tests/conftest.py
```diff
@@ -259,6 +275,80 @@ def mod_cylc_validate_cli(mod_capsys, mod_caplog):
...
+        nonlocal monkeypatch, request, run_dir
+
+        # point rose-stem at the desired run directory
+        rose_stem_opts['no_run_name'] = True
+        rose_stem_opts['workflow_name'] = str(
+            run_dir.relative_to(CYLC_RUN_DIR)
+        )
+
+        # make it look like we're running the "rose stem" CLI
+        monkeypatch.setattr('sys.argv', ['stem'])
```
wxtim, line 298: Why is this not `['rose', 'stem']`?

tests/functional/test_rose_stem.py
wxtim: This whole structure was nasty. I'm glad you've killed it.

My aim at the time I did it was to replicate the old functional tests as closely as possible without adding the TAP framework to Cylc Rose.

wxtim (approved):
Nice: A much stronger test framework than the one I created.

tests/conftest.py
oliver-sanders: Looks like we're on Pytest 8 now so this should be no problem.

tests/conftest.py
oliver-sanders: You tell me!

oliver-sanders: Presumably the Rose root command swallows the first argument?

oliver-sanders:
BTW: please squash merge this when the time comes, I've only put this up as micro commits for ease of review.

MetRonnie (approved):

Issue #319, opened 2024-05-07
compat mode ignored


oliver-sanders:
A bug which can be replicated by:

* Running a Cylc 7 workflow under Cylc 8 compat mode.
* Which contains an fcm_make task.
* And a `:finish` trigger.

This should do it:

```ini
# suite.rc
[scheduling]
  [[graph]]
    R1 = fcm_make:finish => fin
```

Due to the way the workflow config is loaded from the `log/config/flow-processed.cylc` file, the Cylc 7 compat mode check is not being performed correctly causing the config to think it is not in compat mode.

This check is performed automatically by `WorkflowConfig.__init__` so may require an upstream change to cylc-flow, hence 1.4.0.

Introduced by https://github.com/cylc/cylc-rose/pull/302

wxtim:
I cannot replicate the bug with your example.

oliver-sanders:
The error will be raised when the fcm_make task runs.

wxtim:
Do we have an example of the error?

PR #322, opened 2024-05-10
Fix.cylc rose issue 319

wxtim:
Works with https://github.com/cylc/cylc-flow/pull/6097 to close https://github.com/cylc/cylc-rose/issues/319.

Functional test provided by https://github.com/metomi/rose/pull/2779

**Check List**

### DIFF
--- a/cylc/rose/platform_utils.py
+++ b/cylc/rose/platform_utils.py
@@ -23,12 +23,13 @@ from pathlib import Path
 import sqlite3
 import subprocess
 from time import sleep
-from typing import Any, Dict
+from typing import Any, Dict, Union
 
 from cylc.flow.config import WorkflowConfig
 from cylc.flow.id_cli import parse_id
 from cylc.flow.pathutil import (
     get_workflow_run_config_log_dir,
+    get_workflow_run_dir,
     get_workflow_run_pub_db_path,
 )
 from cylc.flow.workflow_files import WorkflowFiles
@@ -62,7 +63,7 @@ def get_platform_from_task_def(flow: str, task: str) -> Dict[str, Any]:
     config = WorkflowConfig(
         flow,
         flow_file, Values(),
-        force_compat_mode=force_compat_mode(flow_file)
+        force_compat_mode=get_compat_mode(get_workflow_run_dir(workflow_id))
     )
     # Get entire task spec to allow Cylc 7 platform from host guessing.
     task_spec = config.pcfg.get(['runtime', task])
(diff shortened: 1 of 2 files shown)


cylc/rose/platform_utils.py
```diff
@@ -56,7 +57,12 @@ def get_platform_from_task_def(flow: str, task: str) -> Dict[str, Any]:
         workflow_id,
         WorkflowFiles.FLOW_FILE_PROCESSED,
     )
-    config = WorkflowConfig(flow, flow_file, Values())
+
+    config = WorkflowConfig(
+        flow,
+        flow_file, Values(),
+        force_compat_mode=force_compat_mode(flow_file)
```
oliver-sanders, line 64: Note, needs matching import.

```suggestion
        force_compat_mode=get_compat_mode(get_workflow_run_dir(workflow_id))
```

cylc/rose/platform_utils.py
```diff
@@ -75,6 +81,15 @@ def get_platform_from_task_def(flow: str, task: str) -> Dict[str, Any]:
...
 
+def force_compat_mode(flow_file):
+    """Check whether this is a Cylc 7 Back compatibility mode workflow:
+    https://github.com/cylc/cylc-rose/issues/319
+    """
+    return (
+        Path(flow_file).parent.parent.parent / WorkflowFiles.SUITE_RC
+    ).exists()
+
+
```
oliver-sanders, line 92: Could simplify this by passing in the run directory rather than walking up the tree of parents to get there.

```suggestion
def check_compat_mode(run_dir):
    """Check whether this is a Cylc 7 Back compatibility mode workflow.
    
    See https://github.com/cylc/cylc-rose/issues/319
    
    Args:
        run_dir: Cylc workflow run directory.

    """
    return (run_dir / WorkflowFiles.SUITE_RC).exists()


```

oliver-sanders (commented):
Looks good, possible simplification.

cylc/rose/platform_utils.py
```diff
@@ -75,6 +81,15 @@ def get_platform_from_task_def(flow: str, task: str) -> Dict[str, Any]:
     return platform
 
 
+def force_compat_mode(flow_file):
```
wxtim, line 84:
```suggestion
def get_compat_mode(workflow_id):
```

oliver-sanders (changes requested):
Note, the issue is tagged against 1.4.0 not 1.3.x. We can leave this on the bugfix milestones, but will need to amend the cylc-flow dependency in setup.cfg to prevent this from being used with an incompatible release of cylc-flow if we were to make another bugfix release before the cylc 8.3 meta-release.

oliver-sanders:
(tests will fail until upstream PR merged)

oliver-sanders (approved):

oliver-sanders:
Could do with a quick rebase now that the sync PR is in.

cylc/rose/platform_utils.py
```diff
@@ -76,6 +83,17 @@ def get_platform_from_task_def(flow: str, task: str) -> Dict[str, Any]:
...
 
+def get_compat_mode(run_dir: Union[str, Path]) -> bool:
+    """Check whether this is a Cylc 7 Back compatibility mode workflow:
+
+    See https://github.com/cylc/cylc-rose/issues/319
+
+    Args:
+        run_dir: Cylc workflow run directory.
+    """
+    return (Path(run_dir) / WorkflowFiles.SUITE_RC).exists()
```
MetRonnie, line 94: Couldn't we just import and use `cylc.flow.workflow_files.check_deprecation`?

wxtim: I guess that we could, but I'd rather not risk change the state of the `cylc.flow.flags.cylc7_back_compat` flag if at all possible: This approach means that information is only really flowing from Cylc flow into Cylc Rose.

MetRonnie:
Kicking tests

MetRonnie (approved):

Issue #325, opened 2024-06-05
flaky tests


oliver-sanders:
The tests appear to have become flaky, maybe ~3 weeks back?!

https://github.com/cylc/cylc-rose/actions/workflows/tests.yml?query=branch%3Amaster

wxtim:
@oliver-sanders: comparing the time of this post and looking at test failures in the week preceeding that: All these common failures are the one logged by Ronnie.

wxtim:
Doesn't seem flaky - just run master 3 times in quick succession: https://github.com/cylc/cylc-rose/actions/workflows/tests.yml

PR #331, opened 2024-06-19
setup: bump upstream component versions

oliver-sanders:


MetRonnie (approved):
Test failures are #325

PR #336, opened 2024-07-19
glbl_cfg: reload without mutating the original

oliver-sanders:
Sibling: https://github.com/cylc/cylc-flow/pull/6249

* Use the new `reload` kwarg rather than calling the `.load()` method.
* Fixes https://github.com/cylc/cylc-flow/issues/6244
* The `.load()` method mutates the existing config, due to the use of logging within (and outside of) this routine and the use of `glbl_cfg` in the logging, this created a race condition.
* The new `reload` kwarg creates a new config instance, then sets this as the default.

**Check List**

- [x] Tests are included (test for the original feature added in the PR - https://github.com/cylc/cylc-rose/pull/269, tests for the bugfix in the cylc-flow PR)
- [x] Changelog entry (added an entry in cylc-flow)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,7 +56,7 @@ python_requires = >=3.7
 include_package_data = True
 install_requires =
     metomi-rose==2.3.*
-    cylc-flow==8.3.*
+    cylc-flow>=8.3.3.dev,<8.4
     metomi-isodatetime
     ansimarkup
     jinja2
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,7 +56,7 @@ python_requires = >=3.7
 include_package_data = True
 install_requires =
     metomi-rose==2.3.*
-    cylc-flow>=8.3.3.dev,<8.4
+    cylc-flow>8.3.2,<8.4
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -56,7 +56,7 @@ python_requires = >=3.7
 include_package_data = True
 install_requires =
     metomi-rose==2.3.*
-    cylc-flow==8.3.*
+    cylc-flow>=8.3.3.dev,<8.4
```
wxtim, line 59: Are we going to release 8.4 without the fix? 
Do we need the upper pin?
Do we want to put the .dev in here, or should we use `8.3.4`?

MetRonnie: > Do we want to put the .dev in here, or should we use `8.3.4`?

See latest commit

> Do we need the upper pin?

Yes, for consistency with previous `8.3.*` pin

hjoliver (approved):

wxtim:
Re-running checks post https://github.com/cylc/cylc-flow/pull/6249

PR #345, opened 2024-10-11
Reinstall picks up installed config.

wxtim:
Closes https://github.com/cylc/cylc-flow/issues/6286

Partner of https://github.com/cylc/cylc-flow/pull/6404, required to close https://github.com/cylc/cylc-flow/issues/6286

Use `options.against_source = rundir` to merge new CLI options with those stored in an earlier install's `rose-suite-cylc-install.conf`.

**Check List**

- [x] Tests are included, and will fail until the companion PR is merged.


tests/unit/test_config_node.py
```diff
@@ -345,3 +347,55 @@ def test_deprecation_warnings(
         assert must_include in records
     else:
         assert must_exclude not in records
+
+
+def test_retrieve_installed_cli_opts(tmp_path):
```
wxtim, line 352: @oliver-sanders - I've not dotted all the i's or crossed all the t's on this PR - but I'm concerned that this is a fairly fundamental failure - can you check that in your opinion this is failing correctly?

oliver-sanders: Hard to say how this unit should behave without a full review.

However, at a functional level, this:

```
$ cylc vr -S KEY=value
```

Should result in this:

```
$ cylc validate -S KEY=value
$ cylc reinstall -S KEY=value
$ cylc {reload,play}
```

I.E. The prior value of `KEY` should make no difference to command outcome.

wxtim: And presumably if I've changed the source dir that should override the old dir?

oliver-sanders: I think it's the same question.

If you `cylc vr -S KEY=value`, then I would expect `cylc reinstall -S KEY=value`. Because CLI options override config options (they have to), the CLI option wins.

wxtim: OK. Just seems weird. But essentially if you set it by CLI you have to modify it in the same way. That's good, because that's what is happening. 😄

oliver-sanders: > What should happen if you do:
>
> $ cylc install -S KEY=1
> $ printf '[template variables]\nKEY=2\n' > rose-suite.conf
> $ cylc config

In this case you would get `KEY=2`, however, if you did `cylc config --against-source` you would get `KEY=1`.

oliver-sanders: There are four things:

1. New CLI opts
2. Old CLI opts
3. New config
4. Old config

Normally, when we perform an action against a source workflow, we don't have access to the old CLI options / config, we are just looking at the source dir. As a result, we only do 1 & 3.

When we use `--against-source` we are reading the source workflow with the context of the installed workflow. This turns on 2 & 3.

cylc/rose/stem.py
```diff
@@ -575,7 +575,7 @@ def get_rose_stem_opts():
             " the groups into a series of tasks to run."
         ),
         action="append",
-        metavar="PATH/TO/FLOW",
+        metavar="STRING",
```
wxtim, line 578: Not directly relevent to the PR, but this was wrong.

cylc/rose/entry_points.py
```diff
@@ -40,7 +41,15 @@ def pre_configure(srcdir: Path, opts: 'Values') -> dict:
         # nothing to do here
         return {}
 
-    # load the Rose config
+    # If we are validating against source, load saved CLI options
+    # from previous install, as saved in the rose-suite-cylc-install.conf
+    if (
+        getattr(opts, 'against_source', False)
```
oliver-sanders, line 47: It doesn't look like the `--against-source` option was ever accessed from cylc-rose, so we ~*don't* have any compatibility issues to worry about which is good news.

cylc/rose/entry_points.py
```diff
@@ -40,7 +41,15 @@ def pre_configure(srcdir: Path, opts: 'Values') -> dict:
...
+        and isinstance(opts.against_source, Path)
+    ):
+        opts = retrieve_installed_cli_opts(srcdir, opts)
```
oliver-sanders, line 50: Should this have a `modify=False`?

oliver-sanders:
LGTM, tested as working, one question (above).

cylc/rose/entry_points.py
wxtim: No - this function is written to be non-destructive (as you'd want the pre-configure plugin to be). When this function calls `record_cylc_install_options` it does pass `modify=False` to that. We haven't a use (at least not yet) for choosing to have a modifying version of this function.

pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+skip-string-normalization = true
+
+
+[tool.isort]
+profile = "black"
+line_length = 79
+force_grid_wrap = 2
+lines_after_imports = 2
+combine_as_imports = true
+force_sort_within_sections = true
```
oliver-sanders, line 23: \n

wxtim:
```suggestion
force_sort_within_sections = true

```

cylc/rose/utilities.py
```diff
@@ -999,3 +1005,61 @@ def copy_config_file(
...
     return True
+
+
+def retrieve_installed_cli_opts(srcdir, opts):
+    """Merge options from previous install, this install and the CLI.
+
+    Allows validation of merged config for pre-configure where the
+    --against-source argument is used in a Cylc script.
+    """
+
```
oliver-sanders, line 1016:
```suggestion
```

cylc/rose/utilities.py
```diff
@@ -999,3 +1005,61 @@ def copy_config_file(
...
+    # Get opt_conf_keys stored in rose-suite-cylc-install.conf
+    opt_conf_keys = cli_config.value.pop('opts').value.split(' ')
+    if any(opt_conf_keys):
+        opts.opt_conf_keys = opt_conf_keys
+
+    # Get --suite-defines/-S
+    # Work out whether user has used "template variables", "jinja2:suite.rc"
+    # or "empy:suite.rc" (There is an assumption that they aren't mixing
+    # them that is not guarded against):
+    for section in SECTIONS:
```
oliver-sanders, line 1043: Do we need to protect against no template processor being used?

```suggestion
    template_variables = []
    for section in SECTIONS:
```

cylc/rose/utilities.py
```diff
@@ -999,3 +1005,61 @@ def copy_config_file(
...
+        if cli_config.value.get(section, False):
+            template_variables = cli_config.value.pop(section)
+            break
+
+    if any(template_variables):
+        opts.rose_template_vars = [
+            f'{k}={v.value}'
+            for k, v in template_variables.value.items()
+            if v.state == ''
```
oliver-sanders, line 1052: Use `rose.config.ConfigNode.STATE_NORMAL` rather than `''`.

cylc/rose/utilities.py
```diff
@@ -999,3 +1005,61 @@ def copy_config_file(
...
+    # them that is not guarded against):
+    for section in SECTIONS:
+        if cli_config.value.get(section, False):
+            template_variables = cli_config.value.pop(section)
+            break
+
+    if any(template_variables):
+        opts.rose_template_vars = [
+            f'{k}={v.value}'
+            for k, v in template_variables.value.items()
```
oliver-sanders, line 1051: Is this a `ConfigNode`? If so, should we use `ConfigNode.walk` rather than iterating over `.value.items()`?

Note, `.walk` can filter out ignored items.

cylc/rose/utilities.py
```diff
@@ -999,3 +1005,61 @@ def copy_config_file(
...
+            if v.state == ''
+        ]
+
+    # Get all other keys (-D):
+    new_defines = []
+    for l1_key, l1_item in cli_config.value.items():
```
oliver-sanders, line 1057: Is this a `ConfigNode`? If so, should we use `ConfigNode.walk` rather than iterating over `.value.items()`?

Note, `.walk` can filter out ignored items.

cylc/rose/utilities.py
```diff
@@ -999,3 +1005,61 @@ def copy_config_file(
...
+        if isinstance(l1_item.value, Dict):
+            for l2_key, l2_item in l1_item.value.items():
```
oliver-sanders, line 1059: Is this a `ConfigNode`? If so, should we use `ConfigNode.walk` rather than iterating over `.value.items()`?

Note, `.walk` can filter out ignored items.

oliver-sanders (commented):
Works as advertised. Couple of small comments:

cylc/rose/utilities.py
wxtim:
```suggestion
            if v.state == rose.config.ConfigNode.STATE_NORMAL
```

cylc/rose/utilities.py
wxtim: No - It's a list.

wxtim:
Need to check whether I need to bump the version number - does this work/can this be made safe with Cylc<8.3.6

oliver-sanders:
As reported by Mark the other day, the "WARNING [env]ROSE_ORIG_HOST will be ignored" messages which appear on this branch are not very helpful:

* The environment variable has not changed so the warning is invalid.
* This is just an internal thing due to the way this env var is set on installation, it's nothing to do with the options the user specified.
* There is nothing the user can do to avoid this warning.

```
WARNING - [env]ROSE_ORIG_HOST=<myhost> from rose-suite.conf will be ignored: ROSE_ORIG_HOST will be:
    <myhost>
WARNING - [template variables]ROSE_ORIG_HOST=<myhost> from rose-suite.conf will be ignored: ROSE_ORIG_HOST will
    be: <myhost>
```

cylc/rose/entry_points.py
```diff
@@ -40,7 +42,23 @@ def pre_configure(srcdir: Path, opts: 'Values') -> dict:
...
+    elif (
+        getattr(opts, 'against_source', False)
+        and not isinstance(opts.against_source, Path)
+    ):
+        LOG.warning(
+            'Validate --against-source will not check Rose Options until'
+            ' after installation. (Upgrade to Cylc ≥ 8.3.6 for this feature).'
+        )
```
oliver-sanders, line 59: If this branch is not compatible with older versions of cylc-flow, then please up the minimum version requirement rather than logging a warning.

wxtim: It's compatible, but the bug remains unfixed.

oliver-sanders: ???

oliver-sanders:
> As reported by Mark the other day

I think you've turned off this warning completely for `cylc vr`? It (now?) doesn't show when it should?:

```console
$ cylc reinstall -S 'ROSE_ORIG_HOST="elephant"' tmp.2EYla85JQJ
...
WARNING - [template variables]ROSE_ORIG_HOST="elephant" from rose-suite.conf will be ignored: ROSE_ORIG_HOST will be:
    cazld000020.spice.sc.metoffice.gov.uk
REINSTALLED tmp.2EYla85JQJ/run2 from /var/tmp/tmp.2EYla85JQJ
Successfully reinstalled.
Run "cylc reload tmp.2EYla85JQJ/run2" to pick up changes.
Done

$ cylc vr -S 'ROSE_ORIG_HOST="elephant"' tmp.2EYla85JQJ
...
REINSTALLED tmp.2EYla85JQJ/run2 from /var/tmp/tmp.2EYla85JQJ
Successfully reinstalled.
$ cylc reload tmp.2EYla85JQJ/run2
```

tests/functional/test_reinstall_overrides.py
```diff
@@ -58,7 +58,7 @@ async def test_reinstall_overrides(
         opts={'rose_template_vars': ['var="CLIinstall"']})
 
     # Play workflow
-    run_ok(f'cylc play --pause {wid}')
+    run_ok(f'cylc play --pause {wid} --host localhost')
```
wxtim, line 61: Unrelated to PR, but this was causing flakiness becuase the test becoms time-sensitve when running cylc play on a remote host.

oliver-sanders (commented):
Hmm, I'm now getting the same warning three times for `cylc vr`:

```console
$ cylc vr rose-generic -S 'ROSE_ORIG_HOST=42'
$ cylc validate --against-source rose-generic/run1
WARNING - rose_template_vars:ROSE_ORIG_HOST=42 from command line args will be ignored: ROSE_ORIG_HOST will be:
    <myhost>
WARNING - [env]ROSE_ORIG_HOST=<myhost> will be ignored: ROSE_ORIG_HOST will be:
    <myhost>
WARNING - [template variables]ROSE_ORIG_HOST=<myhost> will be ignored: ROSE_ORIG_HOST will be:
    <myhost>
```

The first warning is valid.

The second two are bogus, in both cases the value of `ROSE_ORIG_HOST` is unchanged.

cylc/rose/utilities.py
```diff
@@ -999,3 +1003,94 @@ def copy_config_file(
...
+    have been added to CLI opts, and remove them.
+
+    n.b. This doesn't remove the need to check the input rose-suite.conf
+    for these variables.
+    """
+    options = []
+    for section in ['rose_template_vars', 'defines']:
+        if section in opts.__dict__ and opts.__dict__[section]:
+            for item in opts.__dict__[section]:
+                options.append((section, item))
```
oliver-sanders, line 1085: (avoids accessing `__dict__`)

```suggestion
        for item in getattr(opts, section, []):
            options.append((section, item))
```

(could also convert to list comp)

cylc/rose/utilities.py
```diff
@@ -999,3 +1003,94 @@ def copy_config_file(
...
+
+    for (section, item), (var_name, replace) in itertools.product(
+        options, STANDARD_VARS
+    ):
+        if re.match(rf'.*{var_name}=', item):
```
oliver-sanders, line 1090: With this change `ROSE_ORIG_HOST` will match, but `FOO_ROSE_ORIG_HOST` will not:

```suggestion
        if re.match(rf'.*\b{var_name}=', item):
```

cylc/rose/utilities.py
```diff
@@ -999,3 +1003,94 @@ def copy_config_file(
...
+            LOG.warning(
+                f'{section}:{item} from command line args'
+                f' will be ignored: {var_name} will be: {replace}'
+            )
+            opts.__dict__[section].remove(item)
```
oliver-sanders, line 1095:
```suggestion
            getattr(opts, section).remove(item)
```

I wonder if there is the potential for an accidental double removal to cause a traceback here?

If so, we might want to do:

```suggestion
            with suppress(ValueError):
                getattr(opts, section).remove(item)
```

Note1:

Will require matching import.

Note2:

```python
>>> [].remove('x')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: list.remove(x): x not in list
```

oliver-sanders (commented):
Warnings working well :+1:

Small comments.

cylc/rose/utilities.py
```diff
@@ -999,3 +1003,94 @@ def copy_config_file(
...
+    # if this is a validation against source:
+    no_ignore = False
+    if rundir:
+        no_ignore = True
+        for keys, node in cli_config.walk():
+            if keys[-1] == 'ROSE_ORIG_HOST':
+                node.state = cli_config.STATE_SYST_IGNORED
+
+    # Get opt_conf_keys stored in rose-suite-cylc-install.conf
+    opt_conf_keys = cli_config.value.pop('opts').value.split(' ')
```
oliver-sanders, line 1041:
```suggestion
    opt_conf_keys = cli_config.value.pop('opts').value.split()
```

Slightly safer:

```python
>>> 'a  b  c'.split(' ')
['a', '', 'b', '', 'c']
>>> 'a  b  c'.split()
['a', 'b', 'c']
```

cylc/rose/utilities.py
```diff
@@ -999,3 +1003,94 @@ def copy_config_file(
...
+    # Get --suite-defines/-S
+    # Work out whether user has used "template variables", "jinja2:suite.rc"
+    # or "empy:suite.rc" (There is an assumption that they aren't mixing
+    # them that is not guarded against):
+    for section in SECTIONS:
+        if cli_config.value.get(section, False):
+            template_variables = cli_config.value.pop(section)
+            break
+    # Create new -S list.
+    if any(template_variables):
```
oliver-sanders, line 1054: `template_variables` variable possibly undefined?

setup.cfg
```diff
@@ -56,7 +56,7 @@ python_requires = >=3.7
 include_package_data = True
 install_requires =
     metomi-rose==2.3.*
-    cylc-flow>8.3.2,<8.4
+    cylc-flow>8.3.6,<8.4
```
oliver-sanders, line 59: Whoops, just spotted, this would have excluded the 8.3.6 release!

```suggestion
    cylc-flow>8.3.5,<8.4
```

cylc/rose/utilities.py
wxtim: We already protect against this. Do we need to do it here?

wxtim:
Closed and re-opened to kick the tests onward

oliver-sanders:
Upstream PR merged.

oliver-sanders:
@wxtim, I think the Python 3.7 failure is a legit back-compat issue.

tests/conftest.py
```diff
@@ -457,3 +531,26 @@ def timeout_func(func, message, timeout=5):
...
+    def _inner(code_src):
+        nonlocal tmp_path
+        # Set up paths for test:
+        srcpath = tmp_path / 'src'
+        srcpath.mkdir()
+
+        # the files to install are stored in a directory alongside this
+        # test file:
+        datapath = Path(__file__).parent / code_src
+        copytree(datapath, srcpath, dirs_exist_ok=True)
```
oliver-sanders, line 551: Here it is, just a test file issue, no biggie.

wxtim:
Tried `distutils` but this seems to be something Cylc 3.8 and 3.9 were unhappy with. Gone for a really clunky `run(cp)`.

oliver-sanders:
> https://github.com/cylc/cylc-flow/issues/6076

Fine.

oliver-sanders (approved):

PR #353, opened 2024-11-20
Remove Empy Support

wxtim:
Cylc will no longer support Empy from 8.4.0. This commit removes empy functionality and tests from Cylc Rose.

Follow on from https://github.com/cylc/cylc-flow/pull/6476

n.b. Functional Changes in https://github.com/cylc/cylc-rose/pull/353/commits/bcf16294b03b82e09258e145785f773e4d59e65a. Other commit should just be styling things.

**Check List**


tests/functional/test_rose_stem.py
```diff
@@ -188,7 +188,7 @@ async def test_manual_project_override(rose_stem, rose_stem_project):
     check_template_variables(
         {
             "RUN_NAMES":
-                "['earl_grey', 'milk', 'sugar', " "'spoon', 'cup', 'milk']",
+                "['earl_grey', 'milk', 'sugar', 'spoon', 'cup', 'milk']",
```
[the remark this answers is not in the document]
wxtim: No

oliver-sanders (approved):

cylc/rose/__init__.py
```diff
@@ -59,15 +59,15 @@
    ``opts=A B C``, A space limited list of optional configs.
    ``[env]``, "Variables which the cylc-rose plugin will export to the
    environment."
-   ``[template variables]``, "Variables which can be used by Jinja2 or Empy
+   ``[template variables]``, "Variables which can be used by Jinja2 or
```
markgrahamdawson, line 62:
```suggestion
   ``[template variables]``, "Variables which can be used by Jinja2
```

.github/workflows/tests.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         uses: cylc/release-actions/install-cylc-components@v1
         with:
           cylc_flow: true
-          cylc_flow_opts: 'empy'
+          cylc_flow_opts: ''
```
markgrahamdawson, line 41: If I look for this `cylc_flow_opts` variable in cylc-rose code it only exists in the yaml file. Is it needed?

wxtim, line 41:
```suggestion
```

wxtim: Results of https://github.com/cylc/cylc-rose/actions/runs/12049448140 suggest that it is, in fact, necessary.

oliver-sanders: The [default](https://github.com/cylc/release-actions/blob/1edb53c088f836f1ef49d00c608926bf9296b880/install-cylc-components/action.yml#L59-L62) is `'all'` (equivalent to `pip install cylc-flow[all]`) which is unnecessary for these tests, leaving it as `''` (as you have done) is the correct thing to do.

markgrahamdawson (approved):

PR #363, opened 2025-03-04
actions: use mamba to provision Python environment

oliver-sanders:
* Fixes test failures.
* Get Python from conda-forge rather than the OS vendor.
* Fixes Python 3.7 tests (currently broken due to OS support).
* Test with all supported Python versions.
* Requires https://github.com/cylc/release-actions/pull/100

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,13 +15,17 @@ concurrency:
 env:
   FORCE_COLOR: 2
 
+defaults:
+  run:
+    shell: micromamba-shell {0}
+
 jobs:
   test:
     runs-on: ubuntu-latest
     timeout-minutes: 5
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -15,13 +15,17 @@ concurrency:
 env:
   FORCE_COLOR: 2
 
+defaults:
+  run:
+    shell: bash -elo pipefail {0}
```
MetRonnie, line 20: Should this be (like in https://github.com/cylc/cylc-flow/pull/6655)?
```suggestion
    shell: bash -c "exec $CONDA_PREFIX/bin/bash -elo pipefail {0}"
```

oliver-sanders: `$CONDA_PREFIX/bin/bash` points at the `bash` executable installed by Micromamba.

These Python tests don't have such shell requirements so this PR sticks with using the system installed Bash.

MetRonnie (approved):

PR #378, opened 2025-07-30
Bump dev version

MetRonnie:
Workflow: Release stage 2 - auto publish, run: 65

(The auto PR creation failed because we are waiting for https://github.com/actions/runner-images/issues/12648)

### DIFF
--- a/cylc/rose/__init__.py
+++ b/cylc/rose/__init__.py
@@ -232,4 +232,4 @@ configuration file ``~/cylc-run/my_workflow/opt/rose-suite-cylc-install.conf``.
 
 """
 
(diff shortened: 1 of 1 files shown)


cylc/rose/__init__.py
```diff
@@ -232,4 +232,4 @@
 
 """
 
-__version__ = '1.6.0'
+__version__ = '1.6.1.dev'
```
wxtim, line 235: Master should be 1.7.0.dev?

MetRonnie: We haven't created the 1.6.x branch yet

PR #386, opened 2025-09-12
Make `rose_prune` skip platforms that cannot be found

MetRonnie:
Closes https://github.com/metomi/rose/issues/2924

`rose_prune` should skip platforms that cannot be found (e.g. an invalid platform that doesn't exist).

**Check List**

- [x] No dependency changes 
- [x] Tests are included 
- [x] No docs needed

### DIFF
--- a/tests/unit/test_platform_utils.py
+++ b/tests/unit/test_platform_utils.py
@@ -150,14 +150,18 @@ def fake_flow():
     db_file = get_workflow_run_pub_db_path(flow_name)
     with CylcWorkflowDAO(db_file, create_tables=True) as dao:
         conn = dao.connect()
-        conn.execute(
-            r"INSERT INTO task_jobs (cycle, name, submit_num, platform_name)"
-            r"    VALUES"
-            r"        ('1', 'bar', 1, 'localhost'),"
-            r"        ('1', 'baz', 1, 'dairy'),"
-            r"        ('1', 'bar', 2, 'dairy'),"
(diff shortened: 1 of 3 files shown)


tests/unit/test_platform_utils.py
```diff
@@ -150,14 +150,19 @@ def fake_flow():
...
-            r"        ('1', 'baz', 1, 'dairy'),"
-            r"        ('1', 'bar', 2, 'dairy'),"
-            r"        ('2', 'baz', 1, 'milk')"
-        )
+        conn.execute(r"""
+            INSERT INTO task_jobs (
+                cycle, name, submit_num, platform_name, submit_status
+            )
+                VALUES
+                    ('1', 'bar', 1, 'localhost', 0),
```
wxtim, line 158: I think that this test should also include a case where some but not all submissions to a platform have failed:

```suggestion
                    ('1', 'bar', 1, 'localhost', 0),
                    ('1', 'bar', 1, 'turning', 0),
                    ('1', 'bar', 2, 'turning', 1),
```

MetRonnie:
Changed tack, instead of ignoring platforms based on submit failures, ignore them based on PlatformLookupError. Because submit failure may have still led to file install in some cases?

oliver-sanders (approved):
Makes sense, confirmed interface currently only used by rose_prune

wxtim (approved):

PR #393, opened 2025-10-02
Version update

wxtim:


oliver-sanders (approved):
(make sure the 1.7.x branch is split off before this commit)

PR #425, opened 2026-04-23
remove redundant error check

oliver-sanders:
Remove a redundant error check and bump the coverage threshold.

(This is the cylc-rose repository. It provides a cylc-flow plugin for [metomi-rose](github.com/metomi/rose/).)

**Check List**

### DIFF
--- a/cylc/rose/entry_points.py
+++ b/cylc/rose/entry_points.py
@@ -81,8 +81,7 @@ def post_install(srcdir: Path, rundir: str, opts: 'Values') -> bool:
 
     # perform file installation
     config_node = rose_fileinstall(_rundir, opts)
-    if config_node:
-        dump_rose_log(rundir=_rundir, node=config_node)
+    dump_rose_log(rundir=_rundir, node=config_node)
 
     # Having dumped the config we clear rose options
     # as they do not apply after this.
--- a/cylc/rose/fileinstall.py
+++ b/cylc/rose/fileinstall.py
@@ -17,9 +17,9 @@
 """Utilities related to performing Rose file installation."""
 
 import os
-from typing import TYPE_CHECKING, Union
+from typing import TYPE_CHECKING
 
-from cylc.rose.utilities import rose_config_exists, rose_config_tree_loader
+from cylc.rose.utilities import rose_config_tree_loader
 
 if TYPE_CHECKING:
     from pathlib import Path
@@ -30,11 +30,8 @@ if TYPE_CHECKING:
 def rose_fileinstall(
     rundir: 'Path',
     opts: 'Values',
-) -> 'Union[ConfigNode, bool]':
+) -> 'ConfigNode':
     """Call Rose Fileinstall."""
-    if not rose_config_exists(rundir):
-        return False
-
     # Load the config tree
     config_tree = rose_config_tree_loader(rundir, opts)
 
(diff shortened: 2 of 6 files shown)


cylc/rose/entry_points.py
oliver-sanders, line 72: This checks if the Rose config exists...

cylc/rose/fileinstall.py
```diff
@@ -30,11 +30,8 @@
 def rose_fileinstall(
     rundir: 'Path',
     opts: 'Values',
-) -> 'Union[ConfigNode, bool]':
+) -> 'ConfigNode':
     """Call Rose Fileinstall."""
-    if not rose_config_exists(rundir):
```
oliver-sanders, line 35: ...meaning that this subsequent check was redundant.

tests/functional/test_post_install.py
```diff
@@ -302,10 +302,6 @@ def fakenode(_, __):
         'cylc.rose.utilities.rose_config_tree_loader',
         fakenode,
     )
-    monkeypatch.setattr(
-        'cylc.rose.fileinstall.rose_config_exists',
-        lambda x: True,
-    )
```
oliver-sanders, line 308: So this is no longer necessary.

oliver-sanders:
(whomever merges this PR, please add yourself to the CONTRIBUTING list for this repo)

cylc/rose/entry_points.py
```diff
@@ -81,8 +81,7 @@ def post_install(srcdir: Path, rundir: str, opts: 'Values') -> bool:
 
     # perform file installation
     config_node = rose_fileinstall(_rundir, opts)
-    if config_node:
-        dump_rose_log(rundir=_rundir, node=config_node)
+    dump_rose_log(rundir=_rundir, node=config_node)
```
samuel-denton, line 85: Not sure if this works or not.  rose_fileinstall can return a False value, and I think dump_rose_log will error on a False node value.  Having looked at ConfigDumper() it will try to access properties of config_node and will error since there is no protection.

samuel-denton: Ah, so all instances of 'post_install' calls use the same variable for both srcdir and rundir, so the check for srcdir may be sufficient for rundir also:

Image

That being said, is there a reason to remove the if guard?  It took a fair amount of work to track back whether it's needed or not so I think we should at least put something in the docstring explaining why its safe?

samuel-denton: Or just remove one of the 2 variables for clarity if they are always given the same value on call.

samuel-denton:
BTW your branch name having ++ at the end seems to really upset GH when trying to make edits, if I click any file within your branch I get a 404 error until I refresh.  Doesn't seem to happen on other branches so my guess is the URL getting confused.

cylc/rose/entry_points.py
oliver-sanders: > Not sure if this works or not. rose_fileinstall can return a False value

That was changed as part of this PR: https://github.com/cylc/cylc-rose/pull/425/changes#r3247403334

oliver-sanders: > Ah, so all instances of 'post_install' calls use the same variable for both srcdir and rundir, so the check for srcdir may be sufficient for rundir also:

The src and run dirs can differ, but this isn't a problem. Order of events:
1. Check for the presence of `<src-dir>/rose-suite.conf` - or fail.
2. Copy `<src-dir>/rose-suite.conf` to `<run-dir>/rose-suite.conf`.
3. Load the config.

https://github.com/cylc/cylc-rose/blob/a7e34061b79d4175ed71252abadeb82bd856bace/cylc/rose/entry_points.py#L70-L77

samuel-denton (approved):
LGTM
One review ok here or should this have 2?

Scott-Owen-James (approved):

PR #436, opened 2026-08-11
setup: bump rose dep

oliver-sanders:
**Check List**


setup.cfg
```diff
@@ -50,7 +50,7 @@ packages = find_namespace:
 python_requires = >=3.12
 include_package_data = True
 install_requires =
-    metomi-rose==2.7.*
+    metomi-rose==2.8.*
```
MetRonnie, line 53: (Nothing stopping us from supporting `>=2.7.0,<2.9.0.dev`: https://github.com/metomi/rose/issues?q=milestone%3A2.8.0)

MetRonnie (approved):

setup.cfg
oliver-sanders: True, but unnecessary, and if we did introduce a non-backwards compatible change, it wouldn't get picked up in the tests.

PR #439, opened 2026-09-03
replace two-stage install with one-stage install

### LINKED ISSUE #438: actions: replace two-stage install with one-stage install
At present, we `pip install` the dependencies of cylc-rose first, then we subsequently `pip install` cylc-rose itself:

https://github.com/cylc/cylc-rose/blob/afdd7fa6455e89eb33a142144a8407ff5a06792c/.github/workflows/tests.yml#L46-L57

The two-stage process can be problematic as the second-stage can cause things installed by the first-stage to be reinstalled at different versions!

We have now added a `local_source` option to the `install-cylc-components` action which means we can do this installation in a single stage. Here's how that is done in metomi-rose:

https://github.com/metomi/rose/blob/8231c0412742c09017ca5e2dbe101e3230128b8c/.github/workflows/test.yml#L153-L165

And here are the docs for the `local_source` option:

https://github.com/cylc/release-actions/blob/805994c18059d5cddf773bc73d27b4a3ef6e671a/install-cylc-components/action.yml#L26-L34

To close this issue. switch cylc-rose over to the single-stage install.

Scott-Owen-James:
closes #438 

**Check List**


Scott-Owen-James:
Works, tests that are failing are the same that fail on master: https://github.com/cylc/cylc-rose/actions/runs/34121448577/job/101740188459.

Can't figure out how to replicate the -e part, would appreciate if somebody know or if it even matters if we're using codecov.

.github/workflows/tests.yml
```diff
@@ -3,6 +3,20 @@ name: tests
 on:
   pull_request:
   workflow_dispatch:
+    inputs:
+      rose_ref:
+        description: The Rose branch to test against
+        required: true
```
oliver-sanders: This added option doesn't appear to be used?

.github/workflows/tests.yml
```diff
@@ -3,6 +3,20 @@ name: tests
...
+      cylc_rose_ref:
+        description: The Metomi branch to test against
```
oliver-sanders: "Metomi branch"?

.github/workflows/tests.yml
```diff
@@ -3,6 +3,20 @@ name: tests
...
+        required: true
+
+      cylc_rose_tag:
+        description: The cylc-rose branch to test against
+        required: false
```
oliver-sanders: How does this differ from `cylc_rose_ref`?

Scott-Owen-James: That's embarassing, kept making changes to the workflow dispatch, ended up leaving a bunch of stuff in there I shouldn't have and left out stuff. Will correct shortly.

.github/workflows/tests.yml
Scott-Owen-James: Got rid of it.

.github/workflows/tests.yml
Scott-Owen-James: Corrected.

MetRonnie:
> Works, tests that are failing are the same that fail on master

Test failures on master seem to be caused by https://github.com/cylc/cylc-flow/pull/7223#issuecomment-5584361584

Can ignore the failures for this PR so long as the install step works correctly

.github/workflows/tests.yml
```diff
@@ -43,18 +50,18 @@ jobs:
...
           metomi_rose_opts: ''
-
-      - name: Install cylc-rose
-        # need editable install for coverage
-        run: |
-          pip install -e .[tests]
+          cylc_rose: true
+          cylc_rose_opts: ''
+          cylc_rose_repo: ${{ inputs.cylc_rose_repo || 'cylc/cylc-rose' }}
+          local_source: cylc-rose[tests]
```
MetRonnie: The local source is the current directory because that is where `actions/checkout` puts it by default.
```suggestion
          local_source: .[tests]
```

.github/workflows/tests.yml
```diff
@@ -43,18 +50,18 @@ jobs:
...
```
MetRonnie, line 60: To help with debugging, could you add a new step here
```yml
- run: pip list
```
This is something we've gradually added to other repos but have not gotten round to it here.

.github/workflows/tests.yml
Scott-Owen-James: Your suggestion threw a new error about conflicting dependencies, not sure how or why though...

oliver-sanders:
@Scott-Owen-James, the branch this PR was raised against doesn't match the milestone of the issue.

Scott-Owen-James:
> @Scott-Owen-James, the branch this PR was raised against doesn't match the milestone of the issue.

Drat, will rebase it to 1.7.3.

.github/workflows/tests.yml
```diff
@@ -43,18 +50,18 @@ jobs:
...
           metomi_rose_opts: ''
-
-      - name: Install cylc-rose
-        # need editable install for coverage
-        run: |
-          pip install -e .[tests]
+          cylc_rose: true
+          cylc_rose_opts: ''
+          cylc_rose_repo: ${{ inputs.cylc_rose_repo || 'cylc/cylc-rose' }}
+          local_source: .[tests]
```
MetRonnie, line 58: I believe we also need this
```suggestion
          local_source: .[tests]
          editable_installations: true
```
https://github.com/cylc/release-actions/blob/abd55a24c14c29de719251d50e89c894fb57f7e2/install-cylc-components/action.yml#L20-L24

.github/workflows/tests.yml
```diff
@@ -43,18 +50,18 @@ jobs:
...
           metomi_rose: true
           metomi_rose_opts: ''
-
-      - name: Install cylc-rose
-        # need editable install for coverage
-        run: |
-          pip install -e .[tests]
+          cylc_rose: true
+          cylc_rose_opts: ''
+          cylc_rose_repo: ${{ inputs.cylc_rose_repo || 'cylc/cylc-rose' }}
```
MetRonnie: I don't think this should be here. Based on [the action](https://github.com/cylc/release-actions/blob/abd55a24c14c29de719251d50e89c894fb57f7e2/install-cylc-components/action.yml#L153-L164), `local_source` supersedes this.

@oliver-sanders can you check?
```suggestion
```

oliver-sanders: Yeah, that shouldn't be here.

`cylc_rose: true` is telling it to `pip install cylc-rose`. `local_source: xyz` is telling it to `pip install xyz`.

.github/workflows/tests.yml
```diff
@@ -3,6 +3,13 @@ name: tests
 on:
   pull_request:
   workflow_dispatch:
+    inputs:
+      cylc_rose_ref:
```
oliver-sanders, line 7: now unused?

Scott-Owen-James: cylc_rose_ref is still used in branch, can get rid of cylc_rose_repo though.

Scott-Owen-James:
> @Scott-Owen-James, the branch this PR was raised against doesn't match the milestone of the issue.

Can't seem to rebase to 1.7.3, doesn't seem to exist, should I be doing 1.7.x?

Scott-Owen-James:
Rebase to 1.7.x completed..

MetRonnie:
@Scott-Owen-James outstanding comment: https://github.com/cylc/cylc-rose/pull/439/changes#r3957917098

.github/workflows/tests.yml
```diff
@@ -43,18 +47,17 @@ jobs:
...
           metomi_rose: true
           metomi_rose_opts: ''
-
-      - name: Install cylc-rose
-        # need editable install for coverage
-        run: |
-          pip install -e .[tests]
+          local_source: .[tests]
+          editable_installations: true
+          run: pip list
```
MetRonnie: The suggestion was for it to be its own step (not an input to the step above)
```suggestion

        - run: pip list
```

MetRonnie:
There's a syntax error somewhere: https://github.com/cylc/cylc-rose/actions/runs/34824443753

MetRonnie (approved):
:+1:

These are the correct branches for the 1.7.x release: 
image

